PR #13, opened 2022-04-26
MRD-29 Provide dummy response for overview endpoint

jackmaloney:
https://dsdmoj.atlassian.net/browse/MRD-29

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/controller/OverviewControllerTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/controller/OverviewControllerTest.kt
@@ -1,0 +1,20 @@
+package uk.gov.justice.digital.hmpps.makerecalldecisionapi.controller
+
+import com.natpryce.hamkrest.assertion.assertThat
+import com.natpryce.hamkrest.equalTo
+import org.junit.jupiter.api.Test
+import uk.gov.justice.digital.hmpps.makerecalldecisionapi.domain.Crn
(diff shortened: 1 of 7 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/controller/OverviewControllerTest.kt
```diff
@@ -0,0 +1,20 @@
...
+import uk.gov.justice.digital.hmpps.makerecalldecisionapi.domain.FullName
+
+class OverviewControllerTest {
+  private val overviewController = OverviewController()
+
+  private val crn = Crn("12345")
+
+  @Test
+  fun `returns an overview response successfully`() {
+    val response = overviewController.overview(crn)
```
bills7281, line 16: I like the zen/minimalist test that calls the controller method directly!
A curiousity question ~ Do you use RestAssured/WebTestClient in the integration tests? Or do you tend just to keep it all minimalist like you have here?

By the way, this is a non-blocking question :)

bills7281 (commented):
The Auth and Jwt classes will save me some time on my branch! Thanks :)

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/controller/OverviewControllerTest.kt
jackmaloney: We've not got any real integration tests yet, so that is all up for debate. Looking at the other HMPPS projects, I can see they use WebTestClient - we can discuss it further if you want

jonnywyatt (approved):

PR #29, opened 2022-05-04
MRD-2_community-api-poc

bills7281:
Added CommunityApiClient as a POC.


src/main/resources/application.yml
```diff
@@ -21,6 +21,11 @@ spring:
             client-secret: ${system.client.secret:clientsecret}
             authorization-grant-type: client_credentials
             scope: read
+          community-api:
+            provider: hmpps-auth
+            client-id: ${oauth.client.id}
```
dazoakley, line 26: The creds here will need to be the same as the ones above `${system.client.id:make-recall-decision-api}` etc as far as I know

dazoakley:
Have you got any URLs to pass through for the environment config?  These need to go in the `helm_deploy` directory.

Also we need to think about how this would work in the E2E tests - were you planning on a wiremock container or something to simulate the service?

dazoakley:
Oh oh, you'll also need a healthcheck for the community API - can be a separate PR/Ticket, but worth noting so it doesn't get forgotten.

PR #35, opened 2022-05-06
MRD-18_search-by-crn-helm-yaml-addition

bills7281:
added environment variable to helm_deploy files for offender-search-api url.


src/main/resources/application-dev.yml
```diff
@@ -1,4 +1,4 @@
 offender:
   search:
     endpoint:
-      url: https://probation-offender-search-dev.hmpps.service.justice.gov.uk
+      url: ${probation.offender.search.endpoint.url}
```
dazoakley, line 4: You can set this specifically as `https://probation-offender-search-dev.hmpps.service.justice.gov.uk` if you want - then you won't need to set environment variables for local dev, it's only `application.yml` that needs to be environment driven. :)

bills7281: Lol! Of course ... insert face palm here!

bills7281: Done

dazoakley (approved):

PR #36, opened 2022-05-06
MRD-2_community-api-poc: Added CommunityApiClient.

bills7281:
This provides some of the case summary fields as a poc.

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -26,6 +26,11 @@ spring:
             client-id: ${system.client.id:make-recall-decision-api}
             client-secret: ${system.client.secret:clientsecret}
             authorization-grant-type: client_credentials
+          community-api:
+            provider: hmpps-auth
+            client-id: ${oauth.client.id}
+            client-secret: ${oauth.client.secret}
+            authorization-grant-type: client_credentials
         provider:
           hmpps-auth:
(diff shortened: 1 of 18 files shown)


bills7281:
@dazoakley Darren's commments from abandoned PR - 
Have you got any URLs to pass through for the environment config? These need to go in the helm_deploy directory.

Also we need to think about how this would work in the E2E tests - were you planning on a wiremock container or something to simulate the service (in the docker-compose.yml)?

Oh oh, you'll also need a healthcheck for the community API - can be a separate PR/Ticket, but worth noting so it doesn't get forgotten.

src/main/resources/application.yml
```diff
@@ -26,6 +26,11 @@ spring:
             client-id: ${system.client.id:make-recall-decision-api}
             client-secret: ${system.client.secret:clientsecret}
             authorization-grant-type: client_credentials
+          community-api:
+            provider: hmpps-auth
+            client-id: ${oauth.client.id}
```
dazoakley, line 31: Would be best to keep this consistent to the above:

```
client-id: ${system.client.id:make-recall-decision-api}
client-secret: ${system.client.secret:clientsecret}
```

dazoakley (approved):

PR #43, opened 2022-05-10
MRD-43_MRD-120_person-details-and-overview:

bills7281:
1. Implemented person details endpoint
2. Implemented overview endpoint


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/CaseSummaryOverviewResponse.kt
```diff
@@ -0,0 +1,28 @@
...
+  val crn: String?
+)
+
+data class ProbationTeam(
+  val code: String?,
+  val label: String?
+)
+
+data class Offence(
+  val mainOffence: String?,
```
jackmaloney, line 26: what does mainOffence mean in the context of this? Is it a Yes/No or something else?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/communityapi/Conviction.kt
```diff
@@ -30,3 +31,32 @@ data class Custody(
 data class CustodyStatus(
   val code: String
 )
+
+data class Offence @JsonCreator constructor(
```
jackmaloney, line 35: We have 2 x `Offence` objects - one in this class and one in CaseSummaryOverviewResponse.kt, is that correct or can they be combined? When should you use one over the other?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/CaseSummaryOverviewResponse.kt
bills7281: My mistake, I corrected the type to Boolean.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/communityapi/Conviction.kt
bills7281: The Offence object in this class is in the package 'communityapi'  
indicating that it comes from the community api's conviction endpoint. 

There is also a package for responaes/requests to the communityapi's offender-search endpoint.

All domain objects outside of these packages are responses from our make-recall-decisions-api.

I made two copies as one represents the offence object that the community-api serves up.
The other represents the offence object that our api serves.

I made two in as I see each having a different responsibility.
I'm happy to make a change and have the community-api/our api share if you like :)

jackmaloney: Sounds good - guess if it gets confusing we could always append one of those 2 objects with a different name e.g. ` OffenceResponse`

jackmaloney (approved):

PR #112, opened 2022-05-26
MRD-179_address-correction: Added missing field - streetName.

bills7281:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/PersonDetailsService.kt
```diff
@@ -36,8 +37,8 @@ class PersonDetailsService(
         crn = crn
       ),
       currentAddress = CurrentAddress(
-        line1 = formatTwoWordField(addressNumber, buildingName),
-        line2 = activeAddress?.district ?: "",
+        line1 = formatTwoWordField(addressNumber, streetName),
+        line2 = formatTwoWordField(buildingName, activeAddress?.district ?: ""),
```
jonnywyatt, line 41: Would building name be part of line 1, if present? ie in front of the street name

bills7281: As it is, the building name will always be in line 2. Hmm, which do you think is best Jon? I'll incline to your suggestion :)

bills7281: Done, using [building name] [building number] [street name], verified by Ale :)

jonnywyatt (approved):

jackmaloney (approved):

PR #178, opened 2022-06-23
MRD-121_provide-data-for-risk-graphs-content-sections: MAPPA and RoSH data from ARN added to Risk endpoint.

bills7281:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/CommunityApiClient.kt
```diff
@@ -110,6 +111,26 @@ class CommunityApiClient(
     return result
   }
 
+  fun getAllMappaDetails(crn: String): Mono<MappaResponse> {
+    val responseType = object : ParameterizedTypeReference<MappaResponse>() {}
+    return webClient
+      .get()
+      .uri("/secure/offenders/crn/$crn/risk/mappa")
+      .retrieve()
+//      .onStatus(
```
jackmaloney, line 120: Is this supposed to be commented out?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/config/WebClientUserEnhancementConfiguration.kt
```diff
@@ -18,20 +20,86 @@ import org.springframework.security.oauth2.client.endpoint.DefaultClientCredenti
...
 import java.net.URI
+import javax.servlet.Filter
+import javax.servlet.FilterChain
+import javax.servlet.FilterConfig
+import javax.servlet.ServletException
+import javax.servlet.ServletRequest
+import javax.servlet.ServletResponse
+import javax.servlet.http.HttpServletRequest
+
+// TODO move these out
```
jackmaloney, line 41: Is this comment for something in the future or did you mean to do it before this change goes in?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
```diff
@@ -0,0 +1,201 @@
...
+      Mappa(level = "", isNominal = true, lastUpdated = "")
+    }
+  }
+
+  private suspend fun extractNatureOfRisk(riskSummaryResponse: RiskSummaryResponse): NatureOfRisk {
+    return NatureOfRisk(
+      description = riskSummaryResponse.natureOfRisk ?: "",
+      oasysHeading = OasysHeading(
+        number = "10.2",
+        description = "What is the nature of the risk?"
```
jackmaloney, line 77: What do we do with these descriptions? Just wondering if they should be stored in a config file? Or stored on the frontend?

jackmaloney (commented):
Do the values-{env}.yaml files need to be changed to add the ARN API details?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/CommunityApiClient.kt
bills7281: Yes :) 
If there are no mappa details available, we should return an empty object and not block the Risk screen with a 404.
That way, other Risk details which are available can still be seen.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
bills7281: I talked about it with Jon, at the time we decided to hard code them in the backend.
I'm happy to incline to the FE / a config file. Do you have a preference Jack?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/CommunityApiClient.kt
jackmaloney: But why is it commented? Should it not just be deleted or do you plan on adding back in down the line?

bills7281: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
jackmaloney: Ok if you're happy then I'm happy.

jackmaloney: If at some point we start to think about introducing a Welsh language, we might want to think about moving this to the frontend.

PR #201, opened 2022-06-29
MRD-299_bug-fix-for-offender-search-endpoint:

bills7281:
fixed OffenderSearchService to allow handling of restricted/excluded crns for user.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/OffenderSearchService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/OffenderSearchService.kt
@@ -1,12 +1,15 @@
 package uk.gov.justice.digital.hmpps.makerecalldecisionapi.service
 
+import com.microsoft.applicationinsights.core.dependencies.google.gson.Gson
 import org.springframework.beans.factory.annotation.Qualifier
 import org.springframework.stereotype.Service
+import org.springframework.web.reactive.function.client.WebClientResponseException
 import reactor.core.publisher.Mono
 import uk.gov.justice.digital.hmpps.makerecalldecisionapi.client.CommunityApiClient
 import uk.gov.justice.digital.hmpps.makerecalldecisionapi.client.OffenderSearchApiClient
 import uk.gov.justice.digital.hmpps.makerecalldecisionapi.domain.makerecalldecisions.SearchByCrnResponse
 import uk.gov.justice.digital.hmpps.makerecalldecisionapi.domain.ndelius.OffenderSearchByPhraseRequest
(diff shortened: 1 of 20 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/OffenderSearchService.kt
```diff
@@ -16,36 +16,43 @@ class OffenderSearchService(
...
+        var restricted: Boolean? = null
 
-        if (true == userAccessResponse?.userRestricted || true == userAccessResponse?.userExcluded) {
-          excluded = userAccessResponse.userExcluded
-          restricted = userAccessResponse.userRestricted
-        } else {
-          name = "No name available"
+        // Check whether an empty name is genuinely due to a restriction or exclusion
+        if (it.firstName == null && it.surname == null) {
+          if (true == userAccessResponse?.userRestricted || true == userAccessResponse?.userExcluded) {
```
jackmaloney, line 40: This block of code doesn't look quite right now. The block can never be true as the code on line 20 always returns first. 

When I put the original code in for this, I was trying to stop it hitting the `communityApiClient.getUserAccess()` call for every single request, partly because it isn't future proof if we expand search in the future for search terms that are more broad than crn. For example, searching by name could return 100s or 1000s of cases and you would need to do an API call to `getUserAccess()` for every one of these which could have a significant effect on performance.

bills7281: > This block of code doesn't look quite right now. The block can never be true as the code on line 20 always returns 
> first.

That's true!! 

Then again the original code AFAIK couldn't have worked either. The comment says -
` // Check whether an empty name is genuinely due to a restriction or exclusion`
This assumes that the name is retrievable if there is a restriction/exclusion. As far as I can see it isn't.
If there is a restriction an attempt to call to the probation-search API will result in a 403.

I think the solution here is to remove the null name check.

> When I put the original code in for this, I was trying to stop it hitting the communityApiClient.getUserAccess() call for every single request, partly because it isn't future proof

That might be prudent. My opinion here would be to stick to the Extreme Programming YAGNI principle, and cross that bridge when we come to it.

I could be wrong, so correct me if I am! I'm also not married to my proposals, so feel free to disagree!

Hmm, this feels more like a huddle conversation :)
Anyway, we can have a call about it soon!

bills7281: Addressed as discussed.

jackmaloney (approved):

PR #233, opened 2022-07-19
Recommendations Update & activeRecommendation object and active status added to Create Recommendation endpoint

bills7281:
* Added update endpoint for recommendatons and 
* put an activeRecommendation object on the create recommendation endpoint.

PR #235, opened 2022-07-21
MRD-342 Active recommendation response to all endpoints

jackmaloney:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/entity/RecommendationEntity.kt
```diff
@@ -37,7 +42,7 @@ enum class Recommendation(val text: String) {
 }
 
 enum class Status {
-  DRAFT
+  DRAFT, SUBMITTED
```
jonnywyatt, line 45: How about `DOCUMENT_CREATED` rather than `SUBMITTED` (which is quite generic and might be read as 'submitted to delius' or 'submitted to PPCS' which could both be future statuses?)

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationServiceTest.kt
```diff
@@ -230,6 +230,38 @@ internal class RecommendationServiceTest : ServiceTestBase() {
...
+    assertThat(result?.lastModifiedDate).isEqualTo(recommendation.data.lastModifiedDate)
+  }
+
+  @Test
+  fun `get the latest draft recommendation for CRN when multiple draft recommendations exist in database`() {
+    val recommendation1 = RecommendationEntity(id = 1, data = RecommendationModel(crn = crn, lastModifiedBy = "John Smith", lastModifiedDate = "2022-07-19T23:00:00"))
+    val recommendation2 = RecommendationEntity(id = 2, data = RecommendationModel(crn = crn, lastModifiedBy = "Mary Berry", lastModifiedDate = "2022-08-01T10:00:00"))
+    val recommendation3 = RecommendationEntity(id = 3, data = RecommendationModel(crn = crn, lastModifiedBy = "Mary Berry", lastModifiedDate = "2022-08-01T11:00:00"))
+    val recommendation4 = RecommendationEntity(id = 4, data = RecommendationModel(crn = crn, lastModifiedBy = "Mary Berry", lastModifiedDate = "2022-08-01T09:00:00"))
+    val recommendation5 = RecommendationEntity(id = 5, data = RecommendationModel(crn = crn, lastModifiedBy = "Harry Winks", lastModifiedDate = "2022-07-26T12:00:00"))
```
jonnywyatt, line 253: Will the returned date-times in the response have a UTC timezone marker?
eg `2022-07-26T12:00:00.000Z`

jackmaloney: This field was put in by Bill the other day. Checking the database, I can't see the Z timezone marker but might be worth @bills7281 confirming how this works.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/entity/RecommendationEntity.kt
jackmaloney: I don't think 'submitted to Delius' would ever be a state, considering Delius will one day be deprecated. I'd envisage this being a flag on the case if it was needed, especially if the document needed to be submitted to multiple places - you wouldn't want to store this in the status as they'd end up writing over each other. I tried not to put too much thought into this as it kind of depends on frontend design and don't think we've got that far yet. Let me speak to Beck and see what she thinks.

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationServiceTest.kt
jonnywyatt: This change? https://github.com/ministryofjustice/make-recall-decision-api/pull/233#discussion_r924290672

bills7281: Checking now @jackmaloney @jonnywyatt

bills7281: The date given here is mocked "2022-08-01T10:00:00", so it doesn't come from the repo.

The actual date String sent to the DB is'2022-07-21T09:56:21.034083274'. It is UTC zoned.
But it's format is not '2022-07-26T12:00:00.000Z'.

I can make a seperate PR to address this :) @jonnywyatt @jackmaloney

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/entity/RecommendationEntity.kt
jackmaloney: Updated to `DOCUMENT_CREATED` after discussing with Beck

bills7281 (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationServiceTest.kt
jackmaloney: Bill will address this separately

PR #241, opened 2022-07-21
Enrichment of GET Recommendations Endpoint

bills7281:
I've added -

* recallType (not the recall type object but a String eg/NO_RECALL) as specified on https://dsdmoj.atlassian.net/browse/MRD-328
* status eg/ DRAFT and
* activeRecommendations object


src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationServiceTest.kt
```diff
@@ -208,7 +208,9 @@ internal class RecommendationServiceTest : ServiceTestBase() {
 
   @Test
   fun `get a recommendation from the database`() {
-    val recommendation = Optional.of(RecommendationEntity(data = RecommendationModel(crn = crn)))
+    val recommendation = Optional.of(
+      RecommendationEntity(data = RecommendationModel(crn = crn, status = Status.DRAFT, recommendation = Recommendation.NO_RECALL, lastModifiedBy = "Bill", lastModifiedDate = "2022-05-18T19:33:56"))
```
jonnywyatt, line 212: Should the date be UTC? `2022-05-18T19:33:56.000Z`
https://dsdmoj.atlassian.net/wiki/spaces/MRD/pages/4009590838/3.+How+to+store+times+and+adjusting+for+DST

bills7281: The real Entity from the db will be UTC. The mock here doesn't care, any String will do.

jackmaloney (approved):

PR #265, opened 2022-08-03
MRD-436 Add custodyStatus to recommendation data

jackmaloney:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/entity/RecommendationEntity.kt
```diff
@@ -32,17 +33,27 @@ data class RecommendationEntity(
...
+  var custodyStatus: CustodyStatus? = null,
   var status: Status? = null,
   var lastModifiedBy: String? = null,
   var lastModifiedDate: String? = null,
   val createdBy: String? = null,
   val createdDate: String? = null
 ) : Serializable
 
+enum class Recommendation(val text: String) {
+  STANDARD("Standard"), NO_RECALL("No recall"), FIXED_TERM("Fixed term"),
```
jonnywyatt, line 45: I don't know if this impacts you but the labels are going to change slightly in the front end - `Standard recall`, `Fixed term recall`, `No recall`.
Should they be hardcoded here? The source of truth is the UI

bills7281: These aren't needed anymore :)
Let me check out your branch Jack to see if it all still works with these removed.

jackmaloney: Actually, that's a rebase issue - this shouldn't be in there, Bill removed as part of his last PR. I'll fix now...

bills7281: ah okay :)

jonnywyatt: 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/entity/RecommendationEntity.kt
```diff
@@ -43,6 +45,11 @@ enum class Status {
   DRAFT, DOCUMENT_CREATED, DELETED
 }
 
+data class CustodyStatusOption(
```
jonnywyatt, line 48: Minor point, and don't even know if it's the right way to do it in Kotlin, but every recommendation field that choose from options, will use a structure like this. Could use a generic class?

jackmaloney: Yeh good point, I'll look at putting that in

bills7281 (approved):

jonnywyatt (approved):

PR #278, opened 2022-08-09
MRD-472 Recall type field for Part A

jackmaloney:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/RecallType.kt
```diff
@@ -3,6 +3,17 @@ package uk.gov.justice.digital.hmpps.makerecalldecisionapi.domain.makerecalldeci
...
+  val selected: RecallTypeSelectedValue? = null,
+  val allOptions: List<TextValueOption>? = null
 )
+
+data class RecallTypeSelectedValue(
+  val value: RecallTypeValue? = null,
+  val details: String? = null
+)
+
+enum class RecallTypeValue(val displayValue: String) {
```
jonnywyatt, line 15: Wasn't the plan to keep the source of truth for these in the UI? If we duplicate here, there will be two copies of everything including the longer lists (eg alternatives to recall)

bills7281 (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/RecallType.kt
jackmaloney: Don't think so @jonnywyatt as this is specific to the Part A. We're not duplicating anything here, we just need a way to map the values that come from the frontend over to a value that means something in the Part A document. I don't think this would make sense to store this text in the frontend. You can see the mappings in ticket https://dsdmoj.atlassian.net/browse/MRD-465 for info.

jonnywyatt: ah ok 👍

PR #319, opened 2022-08-18
MRD-456 Licence conditions for Part A document

jackmaloney:
Screenshot 2022-08-18 at 16 08 08


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/StandardLicenceConditions.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import uk.gov.justice.digital.hmpps.makerecalldecisionapi.jpa.entity.TextValueOption
+
+data class StandardLicenceConditions(
+  val selected: List<String>? = null,
+  val allOptions: List<TextValueOption>? = null
+)
+
+enum class SelectedStandardLicenceConditions {
+  GOOD_BEHAVIOUR,
```
jonnywyatt, line 11: I think there are going to be 9 of these; maybe worth checking with Andre

bills7281 (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/StandardLicenceConditions.kt
jackmaloney: Confirmed we are going with 7 for now, we can add the other 2 at a later date

bills7281 (approved):

PR #341, opened 2022-09-01
Change recommendation DB ID to secure random number with 2^32 range

bills7281:


jonnywyatt:
@bills7281 does this change the recommendation ID that's passed back to the UI as part of the response? That's currently a simple integer counter which is used in URLs

jonnywyatt (dismissed):

jonnywyatt (approved):

PR #348, opened 2022-09-06
MRD-595 - Enter police custody location

jonnywyatt:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/util/RecommendationToPartADataMapper.kt
```diff
@@ -16,7 +16,7 @@ class RecommendationToPartADataMapper {
   companion object Mapper {
     fun mapRecommendationDataToPartAData(recommendation: RecommendationEntity): PartAData {
       return PartAData(
-        custodyStatus = recommendation.data.custodyStatus?.selected?.partADisplayValue ?: EMPTY_STRING,
+        custodyStatus = ValueWithDetails(recommendation.data.custodyStatus?.selected?.partADisplayValue, recommendation.data.custodyStatus?.details),
```
jackmaloney, line 19: What does it display if the custodyStatus.selected value is empty? Do we not need the `EMPTY_STRING` part?

jonnywyatt: ok, I've added that

jackmaloney (approved):

PR #352, opened 2022-09-08
MRD-587 Part A - Question 5 sentence details

jackmaloney:


src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationServiceTest.kt
```diff
@@ -47,50 +52,56 @@ internal class RecommendationServiceTest : ServiceTestBase() {
...
+      assertThat(recommendationEntity.id).isNotNull()
+      assertThat(recommendationEntity.data.crn).isEqualTo(crn)
+      assertThat(recommendationEntity.data.status).isEqualTo(Status.DRAFT)
+      assertThat(recommendationEntity.data.personOnProbation).isEqualTo(PersonOnProbation(name = "John Smith", firstName = "John", middleNames = "Homer Bart", surname = "Smith", gender = "Male", ethnicity = "Ainu", dateOfBirth = LocalDate.parse("1982-10-24"), croNumber = "123456/04A", mostRecentPrisonerNumber = "G12345", nomsNumber = "A1234CR", pncNumber = "2004/0712343H"))
+      assertThat(recommendationEntity.data.convictionDetail).isEqualTo(
+        ConvictionDetail(
+          indexOffenceDescription = "Robbery (other than armed robbery)",
+          dateOfOriginalOffence = "2022-08-26",
+          dateOfSentence = "2022-04-26",
+          lengthOfSentence = "6 Days",
```
jonnywyatt, line 93: if this gets displayed in the front end in future it would give more flexibility to send the number and the suffix in separate properties (is that how you're receiving them from delius?)

bills7281 (dismissed):

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationServiceTest.kt
jackmaloney: I did think about this, and then though if it is ever displayed it will almost certainly be displayed together anyway i.e. it wouldn't make sense to show "6" or "Days" without the other. I can change if you think it is needed?

jackmaloney: Done

jonnywyatt (approved):

PR #375, opened 2022-09-16
MRD-591 Minor changes to the format of the Part A downloaded date time

jackmaloney:
- Changed the formatting of the date to dd/mm/yyyy
- Time now in the local date time rather than UTC
Screenshot 2022-09-16 at 14 40 03

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/util/DateTimeHelper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/util/DateTimeHelper.kt
@@ -4,6 +4,7 @@ import org.joda.time.DateTime
 import org.joda.time.DateTimeZone
 import org.joda.time.format.DateTimeFormat
 import java.time.LocalDate
+import java.time.LocalDateTime
 import java.time.format.DateTimeFormatter
 
 class DateTimeHelper {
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/util/DateTimeHelper.kt
```diff
@@ -31,10 +36,10 @@ class DateTimeHelper {
       return date?.format(formatter) ?: ""
     }
 
-    fun splitDateTime(dateTime: String?): Pair<String, String> {
-      val parts = dateTime?.split("T")
-
-      return if (parts != null) Pair(parts[0], parts[1].substring(0, 8)) else Pair("", "")
+    fun splitDateTime(dateTime: LocalDateTime?): Pair<String?, String?> {
+      val formatterDate = DateTimeFormatter.ofPattern("dd/MM/yyyy")
+      val formatterTime = DateTimeFormatter.ofPattern("HH:mm:ss")
```
jonnywyatt, line 41: Is there a requirement to show seconds? Seems very specific

jonnywyatt (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/util/DateTimeHelper.kt
jackmaloney: I can add milliseconds as well if we want to be even more specific?

jackmaloney: Seconds removed

bills7281 (approved):

PR #394, opened 2022-09-27
MRD-706: Added business logic to set `noFixedAbode` on … address to true when a no fixed abode postcode is present

bills7281:
.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/PersonDetailsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/PersonDetailsService.kt
@@ -60,7 +60,7 @@ internal class PersonDetailsService(
         line2 = it.district ?: "",
         town = it.town ?: "",
         postcode = it.postcode ?: "",
-        noFixedAbode = it.noFixedAbode ?: false
+        noFixedAbode = isNoFixedAbode(it)
       )
     }
   }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/PersonDetailsService.kt
```diff
@@ -92,6 +92,13 @@ internal class PersonDetailsService(
     )
   }
 
+  private fun isNoFixedAbode(it: uk.gov.justice.digital.hmpps.makerecalldecisionapi.domain.ndelius.Address): Boolean {
+    val postcodeNoWhiteSpace = it.postcode
+      ?.filter { postcodeCharacter -> !postcodeCharacter.isWhitespace() }
+    val noFixedAbode = postcodeNoWhiteSpace == "NF11NF" || it.noFixedAbode == true
```
jonnywyatt, line 98: I wonder if someone would ever enter in in delius as lowercase, might be worth uppercasing the value before doing this check?

jonnywyatt (commented):
just one suggestion

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/PersonDetailsService.kt
bills7281: Neat point :)! Done!

jonnywyatt (approved):

PR #402, opened 2022-09-29
mrd_710 - Moved adddresses, MAPPA, indexOffenceDetails, and mostRecentPrisonerNumber persistence to the download step.

bills7281:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
```diff
@@ -69,14 +67,11 @@ internal class RecommendationService(
             surname = personDetails?.personalDetailsOverview?.surname,
             gender = personDetails?.personalDetailsOverview?.gender,
             ethnicity = personDetails?.personalDetailsOverview?.ethnicity,
-            dateOfBirth = personDetails?.personalDetailsOverview?.dateOfBirth,
-            mappa = riskResponse?.mappa,
-            addresses = personDetails?.addresses
```
jackmaloney, line 74: As discussed, think we should put this back in for the address

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
```diff
@@ -230,23 +226,43 @@ internal class RecommendationService(
...
       return DocumentResponse(
-        fileName = generateDocumentFileName(recommendationEntity.data, "NAT_Recall_Part_A"),
+        fileName = generateDocumentFileName(updatedRecommendation.data, "NAT_Recall_Part_A"),
         fileContents = fileContents
       )
     }
   }
 
-  private fun generateDocumentFileName(recommendation: RecommendationModel, prefix: String): String {
+  suspend fun updateRecommendation(recommendationEntity: RecommendationEntity): RecommendationEntity {
```
jackmaloney, line 244: We now have 2 methods called updateRecommendation in the same class. Is it worth renaming one so we know when we should use one over the other? Or adding a comment so we clear know the purpose of each one? We should probably move so they appear in the same place in the code as well so it is clear there are 2, what you think?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
```diff
@@ -230,23 +226,43 @@ internal class RecommendationService(
...
-  fun generatePartA(recommendationId: Long, username: String?, userEmail: String?): DocumentResponse {
-
+  suspend fun generatePartA(recommendationId: Long, username: String?, userEmail: String?): DocumentResponse {
     val recommendationEntity = updateRecommendation(null, recommendationId, username, userEmail, true)
     val userAccessResponse = recommendationEntity.data.crn?.let { userAccessValidator.checkUserAccess(it) }
     if (userAccessValidator.isUserExcludedOrRestricted(userAccessResponse)) {
       throw UserAccessException(Gson().toJson(userAccessResponse))
     } else {
-      val fileContents = templateReplacementService.generateDocFromRecommendation(recommendationEntity, DocumentType.PART_A_DOCUMENT)
+      val updatedRecommendation = updateRecommendation(recommendationEntity)
```
jackmaloney, line 235: Seems bit strange to be updating the case twice within the same method. Is there any way this can be refactored so there is only 1 call to update the database? Will add abit of an unnecessary overhead leaving it like this

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
bills7281: Done.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
bills7281: I have renamed the one that just adds the extra fields for clarity.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
bills7281: The first update adds the minimal details which will be required to audit that a Part A download has been attempted, AFAIK it has to come before the user access check.
The second update adds the extra fields, this has to come 'after' the user access check.

Given this, there has to be two update functions.

That said I've annotated the function with @Transactional to ensure that the framework will only make a single call and avoid overhead.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
```diff
@@ -229,24 +226,45 @@ internal class RecommendationService(
...
       return DocumentResponse(
-        fileName = generateDocumentFileName(recommendationEntity.data, "NAT_Recall_Part_A"),
+        fileName = generateDocumentFileName(enrichedRecommendation.data, "NAT_Recall_Part_A"),
         fileContents = fileContents
       )
     }
   }
 
-  private fun generateDocumentFileName(recommendation: RecommendationModel, prefix: String): String {
+  suspend fun pathchRecommendationWithExtraData(recommendationEntity: RecommendationEntity): RecommendationEntity {
```
jackmaloney, line 246: typo in patch

bills7281: Done

jackmaloney (approved):

PR #432, opened 2022-10-17
mrd-700_rule-change-to-show-offence-details-when-at-least-one-active-main-offence-present

bills7281:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
```diff
@@ -89,7 +89,9 @@ internal class RiskService(
       return AssessmentInfo(error = extractErrorCode(ex, "risk assessments", crn))
     }
     val latestCompleteAssessment = getLatestCompleteAssessment(assessmentsResponse)
-    val mainActiveCustodialOffenceFromLatestCompleteAssessment = getMainActiveCustodialOffenceFromLatestCompleteAssessment(activeConvictionsFromDelius, latestCompleteAssessment)
+    val mainActiveCustodialOffenceFromLatestCompleteAssessment = if (hideOffenceDetailsWhenNoMatch) {
```
jackmaloney, line 92: why is there different rules between when it comes from the RecommendationService and when it comes from the CaseSummaryOverviewService? Can't see an AC in MRD-700 to cover that so just wanted to make sure this is intended?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
```diff
@@ -137,6 +139,14 @@ internal class RiskService(
     ?.map { latestAssessment?.offence }
     ?.firstOrNull()
 
+  private fun getMainActiveCustodialOffenceFromLatestCompleteAssessmentWhenNoMatch(
```
jackmaloney, line 142: Is there a test to cover this scenario?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
bills7281: It's not written in the AC. Let me dig out some Slack convos I've had with Andre and update the ticket with some references.

jackmaloney (approved):

PR #434, opened 2022-10-18
mrd-700_overview-ac2-fix: Corrected rules to setting offenceDataFromLatestCompleteAssessment flag.

bills7281:


src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskServiceTest.kt
```diff
@@ -254,7 +254,7 @@ internal class RiskServiceTest : ServiceTestBase() {
 
       // then
       assertThat(response?.lastUpdatedDate).isEqualTo("2022-08-26T15:00:08.000Z")
-      assertThat(response?.offenceDataFromLatestCompleteAssessment).isEqualTo(true)
```
jackmaloney, line 257: Just wondering if the test description can be improved for these 2 that you've changed here? The business rules here are quite complicated and it's not clear to me why this needs to be changed from true to false following the change you made in RiskService. I can't see a test to cover when the offenceDataFromLatestCompleteAssessment field should be true, is this not a valid scenario to test for?

bills7281: I opted to rename the variable tested in the assert to give more clarity.

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskServiceTest.kt
```diff
@@ -263,9 +287,10 @@ internal class RiskServiceTest : ServiceTestBase() {
   }
 
   @Test
-  fun `retrieves assessments hideOffenceDetailsWhenNoMatch is false`() {
+  fun `retrieves assessments hideOffenceDetailsWhenNoMatch is false and offences do not match because a later complete assessment exists and the dates and codes do not match`() {
```
jackmaloney, line 290: think you need the word 'when' after 'assessments'

jackmaloney (approved):

PR #461, opened 2022-11-04
mrd-850 `Recommendation Flow Events` endpoint added.

bills7281:


docker-compose-integration-test-postgres.yml
```diff
@@ -12,6 +12,7 @@ services:
       - "./src/main/resources/db/migration/V1_0__RECOMMENDATIONS_TABLE.sql:/docker-entrypoint-initdb.d/1.sql"
       - "./src/main/resources/db/migration/V1_1__RECOMMENDATIONS_TABLE.sql:/docker-entrypoint-initdb.d/2.sql"
       - "./src/main/resources/db/migration/V1_2__RECOMMENDATIONS_TABLE.sql:/docker-entrypoint-initdb.d/3.sql"
+      - "./src/main/resources/db/migration/V1_11__REC_FLOW_EVENT_TABLE.sql:/docker-entrypoint-initdb.d/4.sql"
```
jonnywyatt, line 15: The events aren't just for rec flow, they're for case summary as well.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/controller/RecFlowEventController.kt
```diff
@@ -0,0 +1,37 @@
...
+@RequestMapping(produces = [MediaType.APPLICATION_JSON_VALUE])
+internal class RecFlowEventController(
+  val recFlowEventService: RecFlowEventService
+) {
+  companion object {
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
+
+  @PreAuthorize("hasRole('ROLE_MAKE_RECALL_DECISION')")
+  @PostMapping("/recommendations/events")
```
jonnywyatt, line 28: The events are for more than the recommendations, they start with 'search-by-crn'

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/RecFlowEvent.kt
```diff
@@ -0,0 +1,26 @@
...
+data class RecFlowEvent(
+  val id: String? = null,
+  val crn: String?,
+  val userId: String?,
+  val timeStamp: String?,
+  val eventType: EventType?
+)
+fun RecFlowEvent.toRecFlowEventModel(): RecFlowEventModel {
+  return RecFlowEventModel(
+    crn = crn,
```
jonnywyatt, line 14: I wonder, if we have the recommendation ID, if we should send that as well. Maybe ok without it. What do you think?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/RecFlowEvent.kt
```diff
@@ -0,0 +1,26 @@
...
+    timeStamp = timeStamp,
+    eventType = eventType
+  )
+}
+
+enum class EventType() {
+  SEARCH_RESULT_CLICKED,
+  MAKE_RECOMMENDATION_CLICKED,
+  UPDATE_RECOMMENDATION_CLICKED,
+  DOWNLOAD_PART_A_CLICKED
```
jonnywyatt, line 25: Can we add `DOWNLOAD_NO_RECALL_LETTER_CLICKED`?

src/main/resources/db/migration/V1_11__REC_FLOW_EVENT_TABLE.sql
```diff
@@ -0,0 +1,7 @@
+drop table if exists recflowevents;
+
+CREATE TABLE recflowevents
+(
+    id serial primary key,
+    data jsonb
```
jonnywyatt, line 6: Is it faster to query and join if there are separate columns for each data point?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/entity/RecFlowEventEntity.kt
```diff
@@ -0,0 +1,39 @@
...
+  @Id
+  open var id: Long = abs(SecureRandom().nextInt().toLong()),
+  @Type(type = "jsonb")
+  @Column(columnDefinition = "jsonb")
+  var data: RecFlowEventModel
+)
+data class RecFlowEventModel(
+  val crn: String?,
+  val userId: String?,
+  val timeStamp: String?,
```
jonnywyatt, line 28: Would this be an ISO date-time?

jonnywyatt (commented):
Any chance we can add a GET endpoint, so we can test it works? eg `/events?crn=123`, returning all items for that CRN?

PR #465, opened 2022-11-09
MRD-870 Improve logging around licence condition service

jackmaloney:


bills7281 (approved):

jonnywyatt (commented):
If we send a metric to appinsights we can track it over time

PR #468, opened 2022-11-10
mrd-865 Added validation around recall types.

bills7281:


src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationServiceTest.kt
```diff
@@ -482,6 +486,38 @@ internal class RecommendationServiceTest : ServiceTestBase() {
...
+          selected = RecallTypeSelectedValue(RecallTypeValue.FIXED_TERM),
+          allOptions = listOf(TextValueOption("NO_RECALL"))
+        )
+      )
+      val json = CustomMapper.writeValueAsString(updateRecommendationRequest)
+      val recommendationJsonNode: JsonNode = CustomMapper.readTree(json)
+
+      try {
+        recommendationService.updateRecommendation(recommendationJsonNode, 1L, "Bill", null, false)
+      } catch (e: InvalidRequestException) {
```
jackmaloney, line 514: How does the test know if the exception has been thrown? Do you not need an assertion somewhere?

bills7281: It's implied. The following line asserts that the recommendation repo has no further interactions. If the exception was never thrown, the update would proceed and the test would fail as an interaction would happen. We have used the same test pattern elsewhere. The controller also tests the exception is thrown by checking that a 400 status is returned on an invalid recall type.

jackmaloney: ok

jackmaloney (approved):

PR #473, opened 2022-11-15
MRD-796 Has personal details page been reviewed

jackmaloney:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
```diff
@@ -182,6 +182,25 @@ internal class RecommendationService(
...
+  ): RecommendationEntity {
+    if (updateRecommendationRequest.hasBeenReviewed?.personOnProbation == true) {
+      val data = recommendationEntity.data
+      val personOnProbation = data.personOnProbation
+      return recommendationEntity.copy(
+        data = data.copy(
+          personOnProbation = personOnProbation?.copy(
+            hasBeenReviewed = true
+          ),
+          hasBeenReviewed = null
```
jonnywyatt, line 197: Should it default to `false`?

bills7281: I copy Jon's comment :)

jackmaloney: This should clear the whole 'hasBeenReviewed ' object from the database after each request from the frontend. We don't really want to store this field in 2 different places in the database as could get confusing, hence the reason for setting to null. The 'hasBeenReviewed' under 'personOnProbation' object will only be true or false as stated in ticket.

bills7281 (approved):

PR #486, opened 2022-11-23
Mrd 285 domain event for dntr letter put on hmpps event queue profiles.

bills7281:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/controller/RecommendationController.kt
```diff
@@ -113,11 +113,14 @@ internal class RecommendationController(
   suspend fun generateDntrDocument(
     @PathVariable("recommendationId") recommendationId: Long,
     @RequestBody documentRequestQuery: DocumentRequestQuery,
+    @RequestHeader("X-Feature-Flags") featureFlags: String?
   ): ResponseEntity<DocumentResponse> {
     log.info(normalizeSpace("Generate DNTR document endpoint for recommendation id: $recommendationId"))
 
+    val mapper = jacksonObjectMapper()
```
jackmaloney, line 120: Maybe we could move this block to a reusable class?

jackmaloney:
Might be worth adding a line into RecommendationController test like this https://github.com/ministryofjustice/make-recall-decision-api/blob/main/src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/RecommendationControllerTest.kt#L482

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/controller/RecommendationController.kt
bills7281: Good idea! Done

PR #497, opened 2022-11-29
mrd-920_MAPPA-review: added hasBeenReviewed to MAPPA field

bills7281:


jonnywyatt (commented):
Will the update endpoint allow this property to be set, same as with previous ones?

jackmaloney (approved):

PR #515, opened 2022-12-14
mrd-970-trigger-events: Create a trigger event

bills7281:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/RecallConsidered.kt
```diff
@@ -0,0 +1,7 @@
+package uk.gov.justice.digital.hmpps.makerecalldecisionapi.domain.makerecalldecisions.recommendation
+
+data class RecallConsidered(
+  val createdDate: String?,
+  val userName: String?,
+  val recallConsideredDetail: String?
```
jackmaloney, line 6: Couple of fields missing here from the contract on ticket - id and userId

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/entity/RecommendationEntity.kt
```diff
@@ -100,7 +102,7 @@ data class RecommendationModel(
 ) : Serializable
 
 enum class Status {
-  DRAFT, DOCUMENT_CREATED, DELETED
+  DRAFT, DOCUMENT_CREATED, DELETED, RECALL_CONSIDERED
```
jackmaloney, line 105: Maybe not part of this PR but if `DOCUMENT_CREATED` isn't used in service, should we just remove it?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
```diff
@@ -63,17 +64,28 @@ internal class RecommendationService(
...
-    username: String?
+    username: String?,
+    featureFlags: FeatureFlags?
   ): RecommendationResponse {
     val userAccessResponse = recommendationRequest.crn?.let { userAccessValidator.checkUserAccess(it) }
     if (userAccessValidator.isUserExcludedRestrictedOrNotFound(userAccessResponse)) {
       throw UserAccessException(Gson().toJson(userAccessResponse))
     } else {
       val personDetails = recommendationRequest.crn?.let { personDetailsService.getPersonDetails(it) }
+      val status = if (featureFlags?.flagConsiderRecall == true) Status.RECALL_CONSIDERED else Status.DRAFT
```
jackmaloney, line 75: Don't we need a change on the update endpoint to change the status from `RECALL_CONSIDERED` to `DRAFT` when they click on 'Make recommendation' button after concerns have been raised? Or is that covered by separate ticket?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationServiceTest.kt
```diff
@@ -112,8 +113,12 @@ internal class RecommendationServiceTest : ServiceTestBase() {
       given(recommendationRepository.save(any())).willReturn(recommendationToSave)
       recommendationService = RecommendationService(recommendationRepository, mockPersonDetailService, templateReplacementService, userAccessValidator, convictionService, riskServiceMocked, communityApiClient, mrdEmitterMocked)
 
+      // and
+      val featureFlags = if (recallConsidered == Status.RECALL_CONSIDERED.toString()) FeatureFlags(flagConsiderRecall = true) else null
```
jackmaloney, line 117: Guess this is fine, just not 100% sure if Status.RECALL_CONSIDERED is the correct thing to be doing the equals to here. You're testing the feature flag value which doesn't really have anything to do with the `Status` object. If you want to leave this in, we just need to remember to clean this test up when we come to removing the feature flag in a few weeks time, as it might not be obvious.

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/RecommendationControllerTest.kt
```diff
@@ -42,22 +42,32 @@ class RecommendationControllerTest() : IntegrationTestBase() {
     userAccessAllowed(crn)
     mappaDetailsResponse(crn, category = 1, level = 1)
     allOffenderDetailsResponse(crn)
+
```
jackmaloney, line 45: Does this need 2 versions of this test? One with feature on and one off? For the one that is on, we should probably assert the new fields are being set correctly.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
bills7281: I suppose it will be part of the update work.

PR #549, opened 2023-01-06
mrd-1008_spo-recall-rationale: Added endpoint for SPO to recaord recall decision.

bills7281:


src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/RecommendationControllerTest.kt
```diff
@@ -192,6 +193,222 @@ class RecommendationControllerTest() : IntegrationTestBase() {
...
+        BodyInserters.fromValue(managerRecallDecisionRequest(decision = "RECALL"))
+      )
+      .headers { it.authToken(roles = listOf("ROLE_MAKE_RECALL_DECISION_SPO")) }
+      .exchange()
+      .expectStatus().isOk
+
+    webTestClient.get()
+      .uri("/recommendations/$createdRecommendationId")
+      .headers { it.authToken(roles = listOf("ROLE_MAKE_RECALL_DECISION")) }
+      .exchange()
```
jackmaloney, line 226: Do we need all these 100s of assertions here. In reality, this will be right at the start of process so shouldn't really be many fields set on a case (in the interest of trying to keep the line count of this test file at a manageable level!)

bills7281: Ok, refactored as suggested!

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
```diff
@@ -208,7 +211,40 @@ internal class RecommendationService(
...
+    recommendationId: Long,
+    readableUserName: String?
+  ): RecommendationResponse {
+    validateManagerRecallDecision(jsonRequest)
+    val existingRecommendationEntity = getRecommendationEntityById(recommendationId)
+    val userAccessResponse = existingRecommendationEntity.data.crn?.let { userAccessValidator.checkUserAccess(it) }
+    if (userAccessValidator.isUserExcludedRestrictedOrNotFound(userAccessResponse)) {
+      throw UserAccessException(Gson().toJson(userAccessResponse))
+    } else {
+      val readerForUpdating: ObjectReader = CustomMapper.readerForUpdating(existingRecommendationEntity.data)
```
jackmaloney, line 239: This block here is copied from the `updateAndSaveRecommendation` method. Wonder if it's worth creating a method of the common bits to make it abit cleaner and reduce copied code, WDYT?

bills7281: Sounds good. I did look at that, but Intellij didn't offer any quick auto refactor solution.
Are you ok if I take that task on in my next tciket?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationServiceTest.kt
```diff
@@ -346,7 +349,154 @@ internal class RecommendationServiceTest : ServiceTestBase() {
...
+      // and
+      var recommendationToSave =
+        existingRecommendation.copy(
+          id = existingRecommendation.id,
+          data = RecommendationModel(
+            crn = existingRecommendation.data.crn,
+            managerRecallDecision = updateRecommendationRequest.managerRecallDecision,
+            personOnProbation = updateRecommendationRequest.personOnProbation,
+            recallType = updateRecommendationRequest.recallType,
+            custodyStatus = updateRecommendationRequest.custodyStatus,
```
jackmaloney, line 449: Should this test only be concerned with updating the `managerRecallDecision` rather than all fields?

bills7281: I kept them in to show that the new update doesn't annihilate any pre-existing data and to follow the pattern of using Mockito's `  then(recommendationRepository).should().save(recommendationToSave)` which of course needs all fields.
That said I suppose I could be over testing! I'll make a new commit with a refactor.

jackmaloney (approved):

PR #551, opened 2023-01-10
Mrd 1017 domain event for spo rationale

bills7281:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
```diff
@@ -199,25 +200,33 @@ internal class RecommendationService(
...
+      existingRecommendationEntity.data.managerRecallDecision = updatedRecommendation.managerRecallDecision
         ?.copy(
           createdDate = utcNowDateTimeString(),
           createdBy = readableUserName
         )
 
       val savedRecommendation = recommendationRepository.save(existingRecommendationEntity)
       log.info("recommendation for ${savedRecommendation.data.crn} updated with manager recall decision")
+      if (savedRecommendation.data.managerRecallDecision?.isSentToDelius == true) {
+        log.info("About to send domain event for ${savedRecommendation.data.crn} on manager recall decision made")
```
jackmaloney, line 222: Might be worth adding the recommendation ID to all the log messages in this method, to make it easier to find relevant log messages when all we know is the ID?

bills7281: done

jackmaloney (dismissed):

jackmaloney (approved):

PR #575, opened 2023-01-26
mrd-1089: modified create endpoint to return an existing in progress recommendation if one exists

bills7281:


src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationServiceTest.kt
```diff
@@ -229,6 +233,48 @@ internal class RecommendationServiceTest : ServiceTestBase() {
     }
   }
 
+  @Test
+  fun `create recommendation when one in progress already exists for case`() {
```
jackmaloney, line 237: Is the test name correct here? Shouldn't it be something like "return existing recommendation when one in progress already exists for case and create endpoint hit" or similar

jackmaloney (approved):

PR #629, opened 2023-04-20
mrd-1232_rework-handling-of-management-oversight-domain-event: introduced new flow for spo rationale and deprecated old

bills7281:


fake-community-api/__files/spo-user-access.json
marcus-bcl, line 1: Your service doesn't call Community API any more - do you need these WireMock mappings?

bills7281: We can indeed 👍 ! I'll chop them out!

PR #666, opened 2023-07-07
mrd-1546_spo-user-info-written-in-pp-section-on-part-a: correction to pp details on part a

bills7281:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapper.kt
```diff
@@ -94,8 +94,8 @@ class PartADocumentMapper : RecommendationDataToDocumentMapper() {
       mappa = recommendation.personOnProbation?.mappa,
       lastRecordedAddress = lastRecordedAddress,
       noFixedAbode = noFixedAbode,
-      lastPersonCompletingFormName = recommendation.userNamePartACompletedBy,
-      lastPersonCompletingFormEmail = recommendation.userEmailPartACompletedBy,
+      probationPracticionerName = recommendation.userNamePartACompletedBy,
```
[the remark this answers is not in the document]
robinandrewdevilliers: Architectural decisions are not the subject of PRs. so this should not block this PR.  However, I feel we are moving in the wrong direction here.  When the part A is created, I see that status info is translated into rec doc.  What should be happening, I think, is that status info should be given to the PartA doc mapper along with the rec doc.  This means less dependency on the rec doc structure by the BE, and more dependency on the status structure which is what that structure is for.

robinandrewdevilliers (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapper.kt
```diff
@@ -94,8 +95,7 @@ class PartADocumentMapper : RecommendationDataToDocumentMapper() {
       mappa = recommendation.personOnProbation?.mappa,
       lastRecordedAddress = lastRecordedAddress,
```
DarrenW-HMPPS, line 96: What's the intention here?
There are several lines in this block of code that are doing similar (e.g. 96, 129, 130).

DarrenW-HMPPS (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapper.kt
DarrenW-HMPPS: Changed my mind.

bills7281: Each line sets a different field.
96 sets the lastRecordedAddress
126 sets the counterSignAcoNAme
130 sets the countersignSpoTime

The purpose of the PartADocumentMapper is to take data from the Recommendation journey (living in the status and rec doc tables). And create a DocumentData object which is used to populate the fields in the word doc.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapper.kt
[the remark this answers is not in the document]
bills7281: All done :)

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapper.kt
```diff
@@ -119,17 +119,20 @@ class PartADocumentMapper : RecommendationDataToDocumentMapper() {
       riskToKnownAdult = recommendation.currentRoshForPartA?.riskToKnownAdult?.partADisplayValue,
       riskToStaff = recommendation.currentRoshForPartA?.riskToStaff?.partADisplayValue,
       riskToPrisoners = recommendation.currentRoshForPartA?.riskToPrisoners?.partADisplayValue,
-      countersignAcoEmail = recommendation.countersignAcoEmail,
-      counterSignSpoEmail = recommendation.countersignSpoEmail,
-      countersignSpoName = recommendation.countersignSpoName,
-      countersignSpoTelephone = recommendation.countersignSpoTelephone,
+
+      countersignAcoEmail = metadata.acoCounterSignEmail,
```
DarrenW-HMPPS, line 123: I think it would be nice if these PO/SPO/ACO values were sorted so that they were all together by "person" and then within that group, in the same order. (i.e. we would have the practitioner, the SPO then the ACO and it would always be name, email, telephone etc.). That makes it easier to see what is present and what is not.

DarrenW-HMPPS (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/entity/RecommendationEntity.kt
```diff
@@ -129,14 +126,18 @@ data class RecommendationModel(
...
-  var countersignSpoName: String? = null,
-  var acoCounterSignEmail: String? = null,
-  var spoCounterSignEmail: String? = null,
-  var countersignSpoDateTime: LocalDateTime? = null,
   val countersignAcoExposition: String? = null,
   val countersignAcoTelephone: String? = null,
-  var countersignAcoName: String? = null,
-  var countersignAcoDateTime: LocalDateTime? = null
+
+  var userNamePartACompletedBy: String? = null, // deprecated
```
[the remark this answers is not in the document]
bills7281: We do as the entity has to follow the structure in the json blob in the DB.
If I deleted these, Spring Data would notice that there are old records in the DB still existing with the old fields and error out. 

Ways around that are 
1) a data migration
2) refactor RecommendationEntity to use a JSONObject type or Map<String,String> instead of the current strongly type entity.

What do you think Darren? Shall I go with one of those options or leave it as it is for now?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/DocumentData.kt
```diff
@@ -48,8 +48,8 @@ data class DocumentData(
   val lastRecordedAddress: String? = null,
   val letterAddress: String? = null,
   val noFixedAbode: String? = null,
-  val lastPersonCompletingFormName: String? = null,
-  val lastPersonCompletingFormEmail: String? = null,
+  val probationPractitionerName: String? = null,
+  val probationPractitionerEmail: String? = null,
```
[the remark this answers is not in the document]
bills7281: It is! The original purpose of the field was indeed to record the name of the PP, which before countersigning was always the PP! Now that we have ACOs and SPOs in the journey I've altered the logic to get the PPs details (not the last user in) and renamed the fields accordingly!

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapper.kt
bills7281: Good idea!

PR #672, opened 2023-07-25
MRD-1573 Add search by name and paging

DarrenW-HMPPS:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/controller/OffenderSearchController.kt
```diff
@@ -25,12 +27,37 @@ internal class OffenderSearchController(
   @PreAuthorize("hasRole('ROLE_MAKE_RECALL_DECISION')")
   @GetMapping("/search")
   @Operation(summary = "Returns a list of people on probation based on a given CRN")
-  suspend fun search(
+  @Deprecated("This endpoint has been replaced by the /paged-search endpoint", level = DeprecationLevel.WARNING)
+  suspend fun search(@RequestParam(required = false) crn: String): List<SearchByCrnResponse> {
+    log.info(normalizeSpace("Offender search endpoint hit for CRN: $crn"))
+    val response = offenderSearchService.search(crn = crn, page = 0, pageSize = 10)
+    return response.results.map {
```
bills7281, line 34: Hopefully this comes through as a non-blocking comment! 
In other places (not all) we've used an extension function to map from one object type as another for example [here](https://github.com/ministryofjustice/make-recall-decision-api/blob/763cb2ba4c32a8fb7de0439548e7909f7c258ef9/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/PersonalDetailsOverview.kt#L28).

Anyway, what you have looks fine to me, but I thought I'd mention the fabulous Kotlin extension function, just in case you have never met it before :D

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/GetValueAndHandleWrappedException.kt
```diff
@@ -3,10 +3,9 @@ package uk.gov.justice.digital.hmpps.makerecalldecisionapi.service
 import reactor.core.publisher.Mono
 import uk.gov.justice.digital.hmpps.makerecalldecisionapi.exception.ClientTimeoutException
 
-fun <T : Any> getValueAndHandleWrappedException(mono: Mono<T>?): T? {
-  return try {
-    val value = mono?.block()
-    value ?: value
```
bills7281, line 9: Nicely spotted :)

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/controller/OffenderSearchController.kt
DarrenW-HMPPS: In this particular bit, I didn't want to create additional code because of the method being deprecated.  The less supporting infrastructure, the easier it should be to tidy up.
I think I've adopted the extension method approach somewhere else in this PR, but I haven't done it for all the mapping instances.

PR #675, opened 2023-07-26
MRD-1573 Move paging values to query parameters

DarrenW-HMPPS:


bills7281:
I think my last review mostly stands as the only bits that have changed are the call to the PI team's pageable endpoint and the tests. I think updating all the wiremock stubs must have been painful 😊! I haven't tested them, but I can do it if you like.

Anyway LGTM, so I approve ✅!

bills7281 (approved):

PR #698, opened 2023-08-30
no_jira_major-incident-update-to-allow-list: allowlist update to provide access to hmpps services

bills7281:


DarrenW-HMPPS (changes requested):
Let's go with not adding these new IP addresses and remove the "FiveWells" addresses that are elsewhere in this file.

DarrenW-HMPPS (approved):

PR #785, opened 2023-11-07
Mrd 1973

robinandrewdevilliers:


fake-ppud-automation-api/__files/search.json
```diff
@@ -0,0 +1,12 @@
+{
+  "results": [
+    {
+      "id": "4F6666656E64657269643D313731383138G725H664",
+      "croNumber": "123456/12A",
+      "nomsId": "JG123POE",
```
DarrenW-HMPPS, line 6: This isn't a valid NOMS ID format so the real PPUD Automation API would reject it.
NOMS ID format is ANNNNAA where A is A-Z and N is numeric digit.

That's what we were told and that's what I've implemented as validation.
However, just Googled it and there are some additional formats ("old" numbers from pre-2007).
https://www.criminaljusticehub.org.uk/jargon-buster/noms-number-2/

Best to make the fake data match a valid format, probably the "new" one.  I will update the PPUD API to accept the older formats.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/PpudAutomationApiClient.kt
```diff
@@ -0,0 +1,55 @@
...
+import reactor.core.publisher.Mono
+import uk.gov.justice.digital.hmpps.makerecalldecisionapi.domain.makerecalldecisions.PpudSearchRequest
+import uk.gov.justice.digital.hmpps.makerecalldecisionapi.domain.makerecalldecisions.PpudSearchResponse
+import uk.gov.justice.digital.hmpps.makerecalldecisionapi.exception.ClientTimeoutException
+import java.time.Duration
+import java.util.concurrent.TimeoutException
+
+class PpudAutomationApiClient(
+  private val webClient: WebClient,
+  @Value("\${ndelius.client.timeout}") private val ppudAutomationTimeout: Long,
```
DarrenW-HMPPS, line 18: Probably needs its own timeout rather than reusing the ndelius client

DarrenW-HMPPS (changes requested):

fake-ppud-automation-api/__files/search.json
robinandrewdevilliers: That noms id is pilfered from the other fakes.  Thanks for the input, but there is a pressing need to clear the steel thread stuff.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/PpudAutomationApiClient.kt
robinandrewdevilliers: Tks - Idea hid this from me.  Good catch.

PR #787, opened 2023-11-07
mrd-2097_admin-and-support-roles: user info wiremock stub update for …

bills7281:
…new test users

### DIFF
--- a/fake-delius-integration-api/__files/user-info-admin.json
+++ b/fake-delius-integration-api/__files/user-info-admin.json
@@ -1,6 +1,6 @@
 {
   "username": "Art Major",
-  "email": "abc@123.com",
+  "email": "bat-phone@domain.com",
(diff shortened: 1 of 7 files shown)


fake-delius-integration-api/__files/user-info-admin.json
```diff
@@ -0,0 +1,10 @@
+{
+  "username": "Art Major",
+  "email": "making-recall-decisions-tech@digital.justice.gov.uk",
```
DarrenW-HMPPS, line 3: Probably better to not expose our email address to the internet.
Looks like you have copied this from elsewhere so it's already exposed, but I would suggest the emails are changed to example.com addresses.

DarrenW-HMPPS (changes requested):

DarrenW-HMPPS (approved):

PR #788, opened 2023-11-10
MRD-2085 - prison api search

robinandrewdevilliers:


src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/PrisonApiControllerTest.kt
```diff
@@ -0,0 +1,41 @@
...
+          .body(
+            BodyInserters.fromValue(prisonOffenderSearchRequest(nomsId)),
+          )
+          .headers {
+            (listOf(it.authToken(roles = listOf("ROLE_MAKE_RECALL_DECISION"))))
+          }
+          .exchange()
+          .expectStatus().isOk,
+      )
+      println(response)
```
DarrenW-HMPPS, line 37: Has this been left by mistake?

DarrenW-HMPPS (changes requested):
There are various bits of PPUD Automation API work in here. Are they intentionally mixed in?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/PrisonApiControllerTest.kt
robinandrewdevilliers: yes tks

robinandrewdevilliers:
> There are various bits of PPUD Automation API work in here. Are they intentionally mixed in?

robinandrewdevilliers:
The code coverage issue is due to a timeout method  that can't be covered in integration testing as I would need to reduce the timeout configured in application-test.yaml to 0.  And that would fail the other tests anyway.  The reason its passing in other clients is probably because they represent less than 20% of the lines in the file.

PR #799, opened 2023-11-18
mrd-2089_add-delete-column-to-db: deleted column added to db with bac…

bills7281:
…kwards compatibility

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/RecommendationResponse.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/RecommendationResponse.kt
@@ -82,6 +82,7 @@ data class RecommendationResponse(
   val releaseUnderECSL: Boolean? = null,
   val dateOfRelease: LocalDate? = null,
   val conditionalReleaseDate: LocalDate? = null,
+  val deleted: Boolean? = false,
 )
 
 data class UnderIntegratedOffenderManagement(
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/RecommendationResponse.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/RecommendationResponse.kt
@@ -82,7 +82,7 @@ data class RecommendationResponse(
   val releaseUnderECSL: Boolean? = null,
   val dateOfRelease: LocalDate? = null,
   val conditionalReleaseDate: LocalDate? = null,
-  val deleted: Boolean? = false,
+  val deleted: Boolean = false,
 )
 
 data class UnderIntegratedOffenderManagement(
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/RecommendationsListItem.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/RecommendationsListItem.kt
@@ -11,4 +11,5 @@ data class RecommendationsListItem(
   val status: Status? = null,
   val statuses: List<RecommendationStatusEntity>? = emptyList(),
   val recallType: RecallType? = null,
+  val deleted: Boolean? = false,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/RecommendationsListItem.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/RecommendationsListItem.kt
@@ -11,5 +11,5 @@ data class RecommendationsListItem(
   val status: Status? = null,
   val statuses: List<RecommendationStatusEntity>? = emptyList(),
   val recallType: RecallType? = null,
-  val deleted: Boolean? = false,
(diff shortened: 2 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/RecommendationResponse.kt
```diff
@@ -82,6 +82,7 @@ data class RecommendationResponse(
   val releaseUnderECSL: Boolean? = null,
   val dateOfRelease: LocalDate? = null,
   val conditionalReleaseDate: LocalDate? = null,
+  val deleted: Boolean? = false,
```
DarrenW-HMPPS, line 85: Can we make this non nullable?  If it's null, it probably has to be interpreted as false, so we should make that the case explicitly.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/RecommendationsListItem.kt
```diff
@@ -11,4 +11,5 @@ data class RecommendationsListItem(
   val status: Status? = null,
   val statuses: List<RecommendationStatusEntity>? = emptyList(),
   val recallType: RecallType? = null,
+  val deleted: Boolean? = false,
```
DarrenW-HMPPS, line 14: Same here and everywhere else.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
```diff
@@ -594,7 +597,8 @@ internal class RecommendationService(
     val legacyRecommendationOpen = recommendationEntity.size > 1
     val legacyStatusOpen = recommendationEntity.isNotEmpty() &&
       recommendationStatusRepository.findByRecommendationId(recommendationEntity[0].id)
-        .any { it.active && (it.name != "CLOSED" && it.name != "DELETED") }
+        .filter { it.active }
+        .none { it.name == "CLOSED" || it.name == "DELETED" }
```
DarrenW-HMPPS, line 601: Are you sure this is doing the same thing as the .any?

Could be wrong here, but I think the any is doing:
return true if any of the items are active but not closed and not deleted

and the filter and none combination is doing:
return true **if there are no active items** or there are active items and none of them are closed or deleted

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
```diff
@@ -918,8 +922,8 @@ internal class RecommendationService(
 
   private fun isStatusOpen(it: RecommendationEntity) =
     recommendationStatusRepository.findByRecommendationId(it.id)
-      .any { it.active && (it.name == "BOOK_TO_PPUD" || it.name == "DNTR_DOWNLOADED") }
-      .not()
+      .filter { it.active }
+      .none { (it.name == "BOOK_TO_PPUD" || it.name == "DNTR_DOWNLOADED") }
```
DarrenW-HMPPS, line 926: Similar here I would guess too.
Is it some kind of refactor suggestion?
I think it would be a good idea to make sure that this has some really strong tests around it.

src/main/resources/db/migration/V1_16__DATA_MIGRATION.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE recommendations
```
DarrenW-HMPPS, line 1: Please can we take the opportunity to bump the version of flyway to V10?

DarrenW-HMPPS (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
bills7281: There already is a decent - [paramaterized test](https://github.com/ministryofjustice/make-recall-decision-api/blob/defea20b5524ad07aa1714ad44cae995307d0642/src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/CaseOverviewControllerTest.kt#L23) for this. It passed with both the original logic and the refactored logic here 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
bills7281: see below :)

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/RecommendationsListItem.kt
bills7281: done

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/RecommendationResponse.kt
bills7281: done

src/main/resources/db/migration/V1_16__DATA_MIGRATION.sql
bills7281: done

DarrenW-HMPPS (approved):

PR #837, opened 2023-12-18
no_jira-npe-fix-on-risk-dates: added null handling for risk dates

bills7281:
I've left in a couple parses for date objects in here -
image
and here -
image
I did this where the code needed to compare/convert `LocalDateTime ` to `LocalDate` and I couldn't see how to avoid parsing 🤔.
That said I've used your _default to 1970 solution for nulls_ before parsing, so I think it's safe.


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
```diff
@@ -86,7 +86,9 @@ internal class RiskService(
   suspend fun getAssessmentStatus(crn: String): String {
     val assessmentsResponse = fetchAssessments(crn)
     val latestAssessment =
-      assessmentsResponse.assessments?.sortedBy { LocalDateTime.parse(it.initiationDate).toLocalDate() }?.reversed()
+      assessmentsResponse.assessments
+        ?.filter { !it.initiationDate.isNullOrEmpty() }
+        ?.sortedBy { LocalDateTime.parse(it.initiationDate).toLocalDate() }?.reversed()
```
DarrenW-HMPPS, line 91: We can sort on just the string and get rid of parsing entirely.  ISO formatted dates can be sorted as they are, that is one of the reasons why they are formatted the way they are - year, month, date.  They are indeed sorted lexigraphically, as you mentioned in your summary, but this gives exactly the same result had they been parsed and sorted chronologically as dates.
The more complication we have, the more chance of a bug (e.g. the one this PR is aiming to fix) so let's do the simplest thing that works.

DarrenW-HMPPS: We shouldn't filter out the assessments with null initiation dates.  They will be the latest assessments if we don't have non-null assessments.  What we have to watch for is sort order of null/empty.  It might be that the best idea is to sort on { initiationDate ?: "1970-01-01" }, then we are clear that we are treating a null date as an old date and reverse will behave transparently and correctly.
Although on the subject of reversed - why don't we skip the reverse and just take last instead of first?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
```diff
@@ -192,8 +194,15 @@ internal class RiskService(
...
-    val historicalScores = riskScoresResponse.sortedBy { it.completedDate }.reversed()
+    val latestScores = riskScoresResponse
+      .filter { !it.completedDate.isNullOrEmpty() }
+      .sortedBy { LocalDateTime.parse(it.completedDate) }
+      .reversed()
+      .firstOrNull()
+    val historicalScores = riskScoresResponse
+      .filter { !it.completedDate.isNullOrEmpty() }
+      .sortedBy { LocalDateTime.parse(it.completedDate) }
+      .reversed()
```
DarrenW-HMPPS, line 205: This is doing the same operation twice, duplicating code and processing, and risking getting a different result (either because the code isn't updated correctly or because of non-determinism in the result).  It should be sorted once, then from that list, take the first.

DarrenW-HMPPS (changes requested):

DarrenW-HMPPS:
Would be worth doing a bit of an audit on the whole codebase to check that whenever we do LocalDate.Parse, we aren't passing in a nullable value.  Where we are passing in a null value, we should change it because it's just a bug waiting to happen.  This is not the first time we have had null problems with date parsing so it's worth being proactive.

I'm not suggesting we get rid of parsing and sort on strings across the whole codebase as part of this PR, just that we check for passing nullables in and fix those ones.  When we crack open a file/class/method in the future, that is when we should remove parsing, doing it as we go.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
bills7281: Done 😌

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
bills7281: Corrected 😌

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/util/MrdTextConstants.kt
```diff
@@ -12,5 +12,6 @@ class MrdTextConstants {
     const val SCORE_NOT_APPLICABLE = "NOT_APPLICABLE"
     const val TICK_CHARACTER = 0x2713.toChar().toString()
     const val NO_NAME_AVAILABLE = "No name available"
+    const val DEFAULT_DATE_TIME_FOR_NULL_VALUE = "1970-01-01T12:00:00.000"
```
DarrenW-HMPPS, line 15: Make this 00:00:00.  This is the Unix epoch. It's a special date representing a value of 0 in legacy Unix systems. Every date in that system is expressed as a number of seconds/milliseconds from that 0 point.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
```diff
@@ -86,8 +87,10 @@ internal class RiskService(
   suspend fun getAssessmentStatus(crn: String): String {
     val assessmentsResponse = fetchAssessments(crn)
     val latestAssessment =
-      assessmentsResponse.assessments?.sortedBy { LocalDateTime.parse(it.initiationDate).toLocalDate() }?.reversed()
-        ?.firstOrNull()
+      assessmentsResponse.assessments
+        ?.map { it.copy(initiationDate = it.initiationDate ?: DEFAULT_DATE_TIME_FOR_NULL_VALUE) }
+        ?.sortedBy { it.initiationDate }
+        ?.lastOrNull()
```
DarrenW-HMPPS, line 93: We don't need the map.  Just do:

?.sortedBy { it.initiationDate ?: DEFAULT_DATE_TIME_FOR_NULL_VALUE }

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
```diff
@@ -151,20 +154,22 @@ internal class RiskService(
 
   private fun getLatestCompleteAssessment(assessmentsResponse: AssessmentsResponse) =
     assessmentsResponse.assessments
-      ?.filter { !it.dateCompleted.isNullOrEmpty() }
-      ?.sortedBy { LocalDateTime.parse(it.dateCompleted).toLocalDate() }
-      ?.reversed()
-      ?.firstOrNull { it.assessmentStatus == "COMPLETE" && it.superStatus == "COMPLETE" }
+      ?.map { it.copy(initiationDate = it.dateCompleted ?: DEFAULT_DATE_TIME_FOR_NULL_VALUE) }
+      ?.sortedBy { it.dateCompleted }
+      ?.lastOrNull() { it.assessmentStatus == "COMPLETE" && it.superStatus == "COMPLETE" }
```
DarrenW-HMPPS, line 159: Again, we don't need the map

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
```diff
@@ -151,20 +154,22 @@ internal class RiskService(
...
     latestAssessment: Assessment?,
     mainOffenceDate: LocalDate?,
-  ) = latestAssessment?.offenceDetails?.any {
-    it.offenceDate != null && (LocalDateTime.parse(it.offenceDate).toLocalDate() == mainOffenceDate)
-  } == true
+  ) = latestAssessment
+    ?.offenceDetails
+    ?.map { it.copy(offenceDate = it.offenceDate ?: DEFAULT_DATE_TIME_FOR_NULL_VALUE) }
+    ?.any {
+      LocalDateTime.parse(it.offenceDate).toLocalDate() == mainOffenceDate
```
DarrenW-HMPPS, line 171: Don't need the map
We can do

.any { it.offenceDate == mainOffenceDate }

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
DarrenW-HMPPS: We don't need the map

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
```diff
@@ -237,7 +244,7 @@ internal class RiskService(
       null
     } else {
       PredictorScore(
-        date = LocalDateTime.parse(riskScoreResponse.completedDate).toLocalDate().toString(),
+        date = LocalDateTime.parse(riskScoreResponse.completedDate ?: DEFAULT_DATE_TIME_FOR_NULL_VALUE).toLocalDate().toString(),
```
DarrenW-HMPPS, line 247: Don't parse something then turn it back into a string.  Just take it as it is.  I would think we don't need to worry about the default here, it can just be null.  The default is just we are clear about what we want when we sort - sometimes you might want null to represent something in the future, we want it to represent something old.

DarrenW-HMPPS (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
bills7281: I don't think this will work here. `offenceDate `is a LocalDateTime String, `mainOffenceDate `is LocalDate, so the conversion/mapping is needed. Correct me if I'm wrong :D

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
bills7281: This is done as `completedDate `is LocalDateTime, but the contract with the UI specifies the date here should be in LocalDate format.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/util/MrdTextConstants.kt
bills7281: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
bills7281: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
bills7281: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
DarrenW-HMPPS: OK.  OffenceDate ought to be a Date not a DateTime (not suggesting we change that now though).

I think in that case, it's OK to go back to the original code for this - it checked for offenceDate being null before parsing so wouldn't suffer from problems.

bills7281:
Hey Darren,

About replacing the mapping as you suggested below -

image

It doesn't appear to be working and I'm getting a null pointer.
Can you spot anything I've done wrong there?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
DarrenW-HMPPS: OK

DarrenW-HMPPS:
> Hey Darren,
> 
> About replacing the mapping as you suggested below -
> 
> image
> 
> It doesn't appear to be working and I'm getting a null pointer (test failing). Can you spot anything I've done wrong there?

It's line 244 in RIskService, still parsing a null value by the looks of it.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RiskService.kt
bills7281: Done

DarrenW-HMPPS: It doesn't look like the reversion has come through.

bills7281: Should be fine now

DarrenW-HMPPS (approved):

PR #845, opened 2024-01-04
mrd-2041_edit-update-part-a-data: added ppud stub ref data for police…

bills7281:
… forces

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/entity/RecommendationEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/entity/RecommendationEntity.kt
@@ -75,7 +75,7 @@ fun RecommendationEntity.toRecommendationResponse(): RecommendationResponse {
 data class RecommendationModel(
   val crn: String?,
   var sensitive: Boolean? = null,
-  var ppudRecordPresent: Boolean? = false,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/entity/RecommendationEntity.kt
```diff
@@ -75,7 +75,7 @@ fun RecommendationEntity.toRecommendationResponse(): RecommendationResponse {
 data class RecommendationModel(
   val crn: String?,
   var sensitive: Boolean? = null,
-  var ppudRecordPresent: Boolean? = null,
+  var ppudRecordPresent: Boolean? = false,
```
DarrenW-HMPPS, line 78: If we are initialising this to false, can we make it non nullable?

DarrenW-HMPPS (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/entity/RecommendationEntity.kt
bills7281: Done 👍

DarrenW-HMPPS (approved):

PR #898, opened 2024-02-15
mrd-2293-bulk-close-recommendations: added bulk close task for stale …

bills7281:
…recommendations on a 21 day cycle

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/repository/RecommendationStatusRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/repository/RecommendationStatusRepository.kt
@@ -1,13 +1,28 @@
 package uk.gov.justice.digital.hmpps.makerecalldecisionapi.jpa.repository
 
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Query
 import org.springframework.data.repository.query.Param
 import org.springframework.stereotype.Repository
 import uk.gov.justice.digital.hmpps.makerecalldecisionapi.jpa.entity.RecommendationStatusEntity
+import java.time.LocalDateTime
 
 @Repository
 interface RecommendationStatusRepository : JpaRepository<RecommendationStatusEntity, Long> {
   fun findByRecommendationId(@Param("recommendationId") recommendationId: Long): List<RecommendationStatusEntity>
   fun findByRecommendationIdAndName(recommendationId: Long, name: String?): List<RecommendationStatusEntity>
   fun findByName(name: String?): List<RecommendationStatusEntity>
+
+  @Query(
+    value = "SELECT DISTINCT rs.recommendation_id\n" +
+      "FROM make_recall_decision.public.recommendation_status rs\n" +
+      "LEFT OUTER JOIN make_recall_decision.public.recommendation_status rs2\n" +
+      "  ON rs.recommendation_id = rs2.recommendation_id\n" +
+      "  AND rs2.active = true\n" +
+      "  AND rs2.name IN ('SENT_TO_PPCS', 'REC_CLOSED')\n" +
+      "WHERE rs.created < :thresholdDate\n" +
+      "  AND rs2.recommendation_id IS NULL",
+    nativeQuery = true,
+  )
+  fun findStaleRecommendations(@Param("thresholdDate") thresholdDate: LocalDateTime): List<Long>
 }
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/repository/RecommendationStatusRepository.kt
```diff
@@ -1,13 +1,35 @@
...
 @Repository
 interface RecommendationStatusRepository : JpaRepository<RecommendationStatusEntity, Long> {
   fun findByRecommendationId(@Param("recommendationId") recommendationId: Long): List<RecommendationStatusEntity>
   fun findByRecommendationIdAndName(recommendationId: Long, name: String?): List<RecommendationStatusEntity>
   fun findByName(name: String?): List<RecommendationStatusEntity>
+
+  @Query(
+    value = "SELECT DISTINCT rs.recommendation_id " +
+      "FROM make_recall_decision.public.recommendation_status rs " +
+      "WHERE NOT EXISTS (" +
```
DarrenW-HMPPS, line 19: It doesn't look like we need 2 sub queries here.  Can the first be replaced with:
`WHERE rs.created < :thresholdDate`?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/jpa/repository/RecommendationStatusRepository.kt
```diff
@@ -1,13 +1,35 @@
...
+      "    WHERE rs2.recommendation_id = rs.recommendation_id " +
+      "    AND rs2.created >= :thresholdDate" +
+      ") " +
+      "AND NOT EXISTS (" +
+      "    SELECT 1 " +
+      "    FROM make_recall_decision.public.recommendation_status rs3 " +
+      "    WHERE rs3.recommendation_id = rs.recommendation_id " +
+      "    AND rs3.active = true " +
+      "    AND rs3.name IN ('SENT_TO_PPCS', 'REC_CLOSED')" +
+      ")",
```
DarrenW-HMPPS, line 31: For the second subquery, LEFT OUTER JOIN might be simpler and more efficient.
So for the whole query we had something like:
```
SELECT DISTINCT rs.recommendation_id
FROM make_recall_decision.public.recommendation_status rs
LEFT OUTER JOIN make_recall_decision.public.recommendation_status rs2
  ON rs.recommendation_id = rs2.recommendation_id
  AND rs2.active = true
  AND rs3.name IN ('SENT_TO_PPCS', 'REC_CLOSED')
WHERE rs.created < :thresholdDate
  AND rs2.recommendation_id IS NULL
```

It might be that Postgres automatically converts a NOT EXISTS into a LEFT OUTER JOIN in the query optimizer though so there might not be an efficiency gain.  
If you prefer the NOT EXISTS from a readability/maintainability point of view, stick with that.

bills7281: Hmm, it seems to behave a bit differently. My original query, will return a 'stale' recommendation if 'any' created date is before the threshold date. This one only returns something if 'all' the created dates are before the threshold date.

DarrenW-HMPPS: I think your original query returned a stale recommendation if there were no created dates AFTER the threshold date.

bills7281: Yes! Which means that if there is 'any' created date before/on the threshold date, then the recommendation will be returned as stale. The new query you gave me appears to require 'all' created dates are before the threshold date.

DarrenW-HMPPS: The original query doesn't consider created dates before the threshold date - it is irrelevant whether there are any created dates before. It is only based on whether there are created dates on or after.

I think my query will return an ID for recommendations that meet both the following criteria:
 a status (any status) was created before the threshold date
 there are no active SENT TO PPCS or REC CLOSED statuses

bills7281: Thanks Darren, I've updated the code and used your query

DarrenW-HMPPS (approved):

PR #930, opened 2024-03-04
mrd-2293_bulk-delete-oom-error-fix: optimised query around fetching d…

bills7281:
…eleted recommenmdation


DarrenW-HMPPS (changes requested):
Can we chat about this Bill? It doesn't make any sense to me.

DarrenW-HMPPS (approved):

PR #964, opened 2024-03-26
mrd-2459_bulk-delete-rework-disable-task: disabling bulk delete sched…

bills7281:
…uled task


DarrenW-HMPPS (changes requested):
I think if we want to disable the job, we should probably delete the job code rather than comment these bits out.

bills7281:
I left it in along with its unit test, repository, detail URL endpoint, and integration tests as the next step is to create something very similar to fire some new app insights events for the recommendations that were deleted originally. I think it's easier to repurpose than to delete it all and then start over.

DarrenW-HMPPS (approved):

PR #991, opened 2024-04-16
mrd-2459_bulk-delete-re-enable: re-enabling stale rec clean-up

bills7281:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationsCleanupTask.kt
```diff
@@ -30,12 +28,11 @@ internal class RecommendationsCleanupTask(
     private val log = LoggerFactory.getLogger(this::class.java)
   }
 
-  // TODO re-enable once the correctAppInsightsEventsForPreviouslyDeletedStaleRecommendations(..) task has run
-//  @Scheduled(
-//    timeUnit = TimeUnit.MINUTES,
-//    initialDelayString = "#{ T(java.util.concurrent.ThreadLocalRandom).current().nextInt(10,20) }",
-//    fixedRateString = "#{ T(java.util.concurrent.ThreadLocalRandom).current().nextInt(10,20) }",
-//  )
+  @Scheduled(
```
DarrenW-HMPPS, line 31: We can relax this schedule now. We don't need to run very frequently at all (once a day is fine) so instead of randomising a value between 10 and 20 minutes, we could make a random initial delay of say between 10 and 60 minutes, then a fixed rate of 24 hours.

DarrenW-HMPPS (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationsCleanupTask.kt
bills7281: Done

DarrenW-HMPPS: Can we make the initial delay longer, so that the risk of 2 starting at the same time is avoided?  The transaction does protects us, but when 2 run at the same time and the transaction protection kicks in we get alerts raised. If we make it between 10 and 60, there is only a 1 in 50 chance of collision.

bills7281: Done

DarrenW-HMPPS (approved):

PR #998, opened 2024-04-24
Mrd 2329 integrate with file upload service

bills7281:


src/main/resources/db/migration/V1_27__RECOMMENDATION_DOCUMENT_TABLE.sql
```diff
@@ -0,0 +1,4 @@
+alter table recommendation_document
```
robinandrewdevilliers, line 1: Don't we need to drop the current binary column from this table?

robinandrewdevilliers (changes requested):

src/main/resources/db/migration/V1_27__RECOMMENDATION_DOCUMENT_TABLE.sql
bills7281: This story is just about upload, not fetching/deleting. If I remove the binary column it will break the current implementation.

robinandrewdevilliers (dismissed):

robinandrewdevilliers (approved):

PR #1029, opened 2024-05-21
mrd-2491_api-for-sar-service: added controller for Subject Access Requests

bills7281:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationHistoryService.kt
```diff
@@ -0,0 +1,32 @@
...
+@Service
+class RecommendationHistoryService(
+  private val recommendationHistoryRepository: RecommendationHistoryRepository,
+) : HmppsProbationSubjectAccessRequestService {
+  override fun getProbationContentFor(
+    crn: String,
+    fromDate: LocalDate?,
+    toDate: LocalDate?,
+  ): HmppsSubjectAccessRequestContent? {
+    val recommendationHistory = if (fromDate != null && toDate != null) recommendationHistoryRepository.findByCrn(crn, fromDate, toDate) else emptyList()
```
DarrenW-HMPPS, line 21: That looks a bit dangerous.  This may lead a caller to believe that we hold no data when in fact we might.  
The dates should be mandatory as part of the request, then can't be null and this ambiguity goes away.

If you have no control over the controller and/or request object because they are part of the HMPPS Kotlin library, it might be better to throw an exception here if the dates are null.

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationHistoryServiceTest.kt
```diff
@@ -0,0 +1,57 @@
...
+    // given
+    service = RecommendationHistoryService(recommendationHistoryRepository)
+    given(recommendationHistoryRepository.findByCrn(any(), any(), any())).willReturn(emptyList())
+
+    // when
+    val result = service.getProbationContentFor("bla", LocalDate.now(), LocalDate.now())
+
+    // then
+    assertThat(result).isNull()
+  }
```
DarrenW-HMPPS, line 35: Would it be better to return empty content rather than null? Or perhaps an object that with an empty list of recommendations but the CRN populated?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationHistoryServiceTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    // when
+    val result = service.getProbationContentFor("bla", LocalDate.now(), LocalDate.now())
+
+    // then
+    assertThat(result?.content).extracting("recommendationId").isEqualTo(123L)
+    assertThat(result?.content).extracting("crn").isEqualTo("123")
+    assertThat(result?.content).extracting("recommendations").asList().first().extracting("createdBy").isEqualTo("Bill")
+  }
+}
```
DarrenW-HMPPS, line 57: It doesn't look like any of this is accessible.  I would have thought that you would have had to implement some sort of service that the controller depends on - a bit like the health check approach.  We don't have health check controllers in our service but we implement the services to provide the health check info.

DarrenW-HMPPS (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationHistoryService.kt
bills7281: >you have no control over the controller and/or request object because they are part of the HMPPS Kotlin library
 That's exactly it Darren 👍! Anyway, I've gone for the exception approach you suggested.

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationHistoryServiceTest.kt
bills7281: Done!

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationHistoryServiceTest.kt
bills7281: >I would have thought that you would have had to implement some sort of service that the controller depends on

Yes, I do (see the interface in the image below). The content type is 'Any' which is why typing isn't accessible here.
image

DarrenW-HMPPS: Is there another PR then? What implements the HmppsProbationSubjectAccessRequestService?

DarrenW-HMPPS (approved):

PR #1054, opened 2024-06-18
Reintroducing the index without concurrently

MancRickster:
Reintroducing the rec_hist index without CONCURRENTLY


andrewl (approved):
approved

PR #1055, opened 2024-06-26
Supressing CVE-2015-6420 since it's specific to some Cisco software

MancRickster:
maven/com.deepoove/poi-tl
is our DocX generating package which references dependency
org.apache.commons/commons-collections4/4.4/

We are using the latest (non-beta) version of poi-tl (1.12.2) and the latest version of org.apache.commons/commons-collections4/ (4.4)


andrewl (approved):
lgtm

PR #1062, opened 2024-07-19
Ensure the retry logic works - currently if the main rec doc update f…

MancRickster:
…unction errors it squelches the exception and rebadges it, which will prevent the retry


LucasCairns (approved):
LGTM 👍

PR #1066, opened 2024-08-16
Mrd 2554 failure retries could result in lost data

MancRickster:
Switch to Jackson and ensuring input param refs aren't changed for possible failure retries.
Reordering domain events to ensure they're only fired after a successful update, and not per retry.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
@@ -3,7 +3,7 @@ package uk.gov.justice.digital.hmpps.makerecalldecisionapi.service
 import com.fasterxml.jackson.databind.JsonNode
 import com.fasterxml.jackson.databind.ObjectReader
 import com.fasterxml.jackson.databind.node.ObjectNode
-import org.json.JSONObject
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.context.annotation.Lazy
@@ -341,21 +341,25 @@ internal class RecommendationService(
     featureFlags: FeatureFlags?,
   ): RecommendationResponse {
     validateRecallType(jsonRequest)
-    val requestJson = JSONObject(jsonRequest.toString())
+    val requestJson = jacksonObjectMapper().readTree(jsonRequest.toString())
     val existingRecommendationEntity = getRecommendationEntityById(recommendationId)
 
-    val sendConsiderationRationaleToDelius = requestJson.optBoolean("sendConsiderationRationaleToDelius")
-    val considerationSensitive = requestJson.optBoolean("considerationSensitive")
+    val sendConsiderationRationaleToDelius = requestJson.has("sendConsiderationRationaleToDelius") &&
+      requestJson.get("sendConsiderationRationaleToDelius").asBoolean()
+    val considerationSensitive = requestJson.has("considerationSensitive") &&
+      requestJson.get("considerationSensitive").asBoolean()
 
-    // remove these from the incoming request as they are not part of the recommendation model, but rather values applies to sending messages to nDelius.
-    val incoming = jsonRequest as ObjectNode
+    // Remove these from the incoming request as they are not part of the recommendation model, but rather values applies to sending messages to nDelius.
+    // Ensure we don't change the 'requestJson' input variable reference otherwise this will affect any possible failure retries
+    val incoming = requestJson.deepCopy() as ObjectNode
     incoming.remove("sendConsiderationRationaleToDelius")
     incoming.remove("considerationSensitive")
 
     val updatedRecommendation: RecommendationModel =
       recommendationFromRequest(existingRecommendationEntity, incoming)
-
-    if (requestJson.optBoolean("sendSpoRationaleToDelius")) {
+    val sendManagementOversightDomainEvent = requestJson.has("sendSpoRationaleToDelius") &&
+      requestJson.get("sendSpoRationaleToDelius").asBoolean()
+    if (sendManagementOversightDomainEvent) {
       log.info("send spo rationale to delius")
       existingRecommendationEntity.data =
         addSpoRationale(existingRecommendationEntity, updatedRecommendation, readableUserName)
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
```diff
@@ -341,30 +341,31 @@ internal class RecommendationService(
     featureFlags: FeatureFlags?,
   ): RecommendationResponse {
     validateRecallType(jsonRequest)
-    val requestJson = JSONObject(jsonRequest.toString())
+    val requestJson = jacksonObjectMapper().readTree(jsonRequest.toString())
     val existingRecommendationEntity = getRecommendationEntityById(recommendationId)
 
-    val sendConsiderationRationaleToDelius = requestJson.optBoolean("sendConsiderationRationaleToDelius")
-    val considerationSensitive = requestJson.optBoolean("considerationSensitive")
+    val sendConsiderationRationaleToDelius = requestJson.has("sendConsiderationRationaleToDelius") &&
```
LucasCairns, line 347: You may be safe with just:
```kotlin
val sendConsiderationRationaleToDelius = requestJson.get("sendConsiderationRationaleToDelius").asBoolean()
```
Looks like `get` won't throw if the property doesn't exist and `asBoolean` will safely convert to `Boolean` defaulting to `false`

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
```diff
@@ -402,11 +398,26 @@ internal class RecommendationService(
     refreshData(pageRefreshIds, existingRecommendationEntity.data)
 
     val result = updateAndSaveRecommendation(existingRecommendationEntity, userId, readableUserName)
```
LucasCairns, line 400: You can avoid creating the `result` variable here by using the `also` [scope function](https://kotlinlang.org/docs/scope-functions.html)

```kotlin
return updateAndSaveRecommendation(existingRecommendationEntity, userId, readableUserName).also {
  log.info("recommendation for ${it.crn} updated for recommendationId $recommendationId")

  if (sendManagementOversightDomainEvent) {
    log.info("send spo rationale to delius")
    sendManagementOversightDomainEvent(recommendationId, existingRecommendationEntity, userId)
  }

  if (sendConsiderationRationaleToDelius) {
    log.info("send consideration rationale to delius")
    sendConsiderationRationaleDomainEvent(recommendationId, existingRecommendationEntity, userId)
  }

  if (sendSpoDeleteRationaleToDelius) {
    log.info("send spo delete recommendation rationale to delius")
    sendDeleteRecommendationDomainEvent(recommendationId, existingRecommendationEntity, userId)
  }

  val sendRecommendationStartedDomainEvent =
    featureFlags?.flagDomainEventRecommendationStarted == true && featureFlags.flagConsiderRecall == true &&
        existingRecommendationEntity.data.recommendationStartedDomainEventSent != true

  if (sendRecommendationStartedDomainEvent) {
    log.info("About to send domain event for ${existingRecommendationEntity.data.crn} on Recommendation started for recommendation id $recommendationId")
    sendRecommendationStartedEvent(existingRecommendationEntity.data.crn)
    log.info("Sent domain event for ${existingRecommendationEntity.data.crn} on Recommendation started asynchronously for recommendation id $recommendationId")
    existingRecommendationEntity.data.recommendationStartedDomainEventSent = true
  }
}
```

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
```diff
@@ -341,30 +341,31 @@ internal class RecommendationService(
...
+    val sendConsiderationRationaleToDelius = requestJson.has("sendConsiderationRationaleToDelius") &&
+      requestJson.get("sendConsiderationRationaleToDelius").asBoolean()
+    val considerationSensitive = requestJson.has("considerationSensitive") &&
+      requestJson.get("considerationSensitive").asBoolean()
 
-    // remove these from the incoming request as they are not part of the recommendation model, but rather values applies to sending messages to nDelius.
-    val incoming = jsonRequest as ObjectNode
+    // Remove these from the incoming request as they are not part of the recommendation model, but rather values applies to sending messages to nDelius.
+    // Ensure we don't change the 'requestJson' input variable reference otherwise this will affect any possible failure retries
+    val incoming = requestJson.deepCopy() as ObjectNode
```
LucasCairns, line 354: Reading up on `deepCopy` looks like this *should* do the trick and avoid mutation of even deeply nested properties of the parameter object

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
```diff
@@ -374,23 +375,18 @@ internal class RecommendationService(
...
-      log.info("send spo delete recommendation rationale to delius")
       existingRecommendationEntity.data = existingRecommendationEntity.data.copy(
         spoDeleteRecommendationRationale = updatedRecommendation.spoDeleteRecommendationRationale,
         sendSpoDeleteRationaleToDelius = updatedRecommendation.sendSpoDeleteRationaleToDelius,
       )
-      sendDeleteRecommendationDomainEvent(recommendationId, existingRecommendationEntity, userId)
     }
 
-    existingRecommendationEntity.deleted = requestJson.has("status") && requestJson.getString("status") == "DELETED"
+    existingRecommendationEntity.deleted = requestJson.has("status") && requestJson.get("status").asText() == "DELETED"
```
LucasCairns, line 389: Cool trick here when modifying multiple fields of an object, you can use the `apply` [scope function](https://kotlinlang.org/docs/scope-functions.html) to avoid repetition

```kotlin
    existingRecommendationEntity.apply {
      deleted = requestJson.has("status") && requestJson.get("status").asText() == "DELETED"
      data.recallConsideredList = updateRecallConsideredList(
        updatedRecommendation,
        data,
        userId,
        readableUserName,
      )
      data = updatePageReviewedValues(updatedRecommendation, existingRecommendationEntity).data
      refreshData(pageRefreshIds, data)
    }
```

LucasCairns (dismissed):
LGTM 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
MancRickster: It's true that `get` wont throw an exception on a null/non existent field, but it could return null, and so we couldn't call `asBoolean` on it without getting a null exception

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
MancRickster: Nice, I like.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
MancRickster: Hmm, this is interesting but I don't really like the look of it, and its readability.

LucasCairns (approved):
LGTM 👍

PR #1080, opened 2024-09-24
Mrd 2563 replace add doc incorrectly handling changes

MancRickster:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/SupportingDocumentService.kt
```diff
@@ -71,27 +71,20 @@ internal class SupportingDocumentService(
     flags: FeatureFlags,
   ) {
     val file = recommendationDocumentRepository.findById(id).orElseThrow { NotFoundException("Supporting document not found") }
-    val crn = recommendationRepository.findById(file.recommendationId).getOrNull()?.data?.crn
+    val crn = recommendationRepository.findById(file.recommendationId!!).getOrNull()?.data?.crn
```
LucasCairns, line 74: `!!` will throw if `recommendationId` is `null` here, you could write it as

```kotlin
  val crn = file.recommendationId
    ?.run {recommendationRepository.findById}
    ?.getOrNull()?.data.crn
```

This will only run if `file.recommendationId` is not null, you could put a `?: throw SomeBadRequestException()` at the end so it can be handled by controller advice 🤔

MancRickster: file.recommendationId can't be null at this point in the workflow (although annoyingly it's nullable in the DB for some reason). I think you're right though, the code suggests null CRN should be catered for, so it's right to maintain that logic. Code amended, ta

LucasCairns (approved):
LGTM 👍

PR #1091, opened 2024-10-04
mrd-2559 ensure correct status codes returned and optional date range for hist…

MancRickster:
…ory/SAR endpoint


LucasCairns (approved):
LGTM 👍

PR #1104, opened 2024-11-08
increase ppud timeout (and CaR timeouts since ppud is proxied through…

MancRickster:
… CaR) to cope with large ppud GetOffender durations


Thomas-Geraghty:
Approving but short term solution, need proper solution 😄

Thomas-Geraghty (approved):

PR #1110, opened 2024-11-19
Mrd 2616 early nomis prisoner offender rec retrieval

MancRickster:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/RecommendationService.kt
```diff
@@ -120,6 +124,13 @@ internal class RecommendationService(
         log.info("Sent domain event for ${recommendationRequest.crn} on Recommendation started asynchronously")
       }
       val personDetails = recommendationRequest.crn?.let { personDetailsService.getPersonDetails(it) }
+      var nomisOffender: Offender? = null
```
LucasCairns, line 127: you _might_ be able to avoid the `var` here by using a try/catch expression

```kt
val nomisOffender: Offender? = try {
  personDetails?.personalDetailsOverview?.nomsNumber?.let { prisonerApiService.searchPrisonApi(it) }
} catch (_: NotFoundException) {
  null.also { log.info("No matching Offender with nomsNumber ${personDetails?.personalDetailsOverview?.nomsNumber} found") }
}
```

MancRickster: nice one, thanks, good spot

ppmg (approved):

PR #1116, opened 2024-12-10
Mrd 2522 remove releases obj from get offender call

MancRickster:
PPUD Offender's Sentences' Releases are no longer returned from getOffender details call, however the 'Release Date' from the Sentence page is (mrd 2515 - return sentence release date from ppud).
Also, some additional test stubs


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/PpudOffender.kt
```diff
@@ -27,6 +30,9 @@ data class PpudSentence(
   val mappaLevel: String,
   val licenceExpiryDate: String?,
   val offence: PpudOffence?,
+  val releaseDate: String?,
+  // deprecated
+  @JsonIgnore
   val releases: List<PpudRelease>?,
```
ppmg, line 36: Why keep the field? Are we still using it somewhere?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/PpudOffender.kt
```diff
@@ -37,6 +43,8 @@ data class PpudOffence(
   val dateOfIndexOffence: String?,
 )
 
+// deprecated
+@JsonIgnoreType
 data class PpudRelease(
```
ppmg, line 48: Same here - do we need to keep the type?

fake-ppud-automation-api/mappings/search-PPCS_X098092.json
ppmg, line 1: Won't this clash with the existing search-PPCS_X738925.json? They have the same pattern matcher.

fake-ppud-automation-api/__files/get-offender-X098095.json
```diff
@@ -32,19 +32,6 @@
           "indexOffence": "Blasphemy against coca cola",
           "dateOfIndexOffence": "2008-06-12"
         },
-        "releases": [
-          {
-            "category": "Not Applicable",
-            "dateOfRelease": "2013-02-02",
```
ppmg, line 38: Should we not be keeping this as the new `releaseDate` field? Or do you deliberately want it not to have one?

fake-ppud-automation-api/__files/get-offender-X098092.json
```diff
@@ -32,6 +32,7 @@
           "indexOffence": "Blasphemy against coca cola",
           "dateOfIndexOffence": "2008-06-12"
         },
+        "releaseDate": "2005-05-12",
```
ppmg, line 35: Should this be `2013-02-02` as it is in the existing release? Or are you changing it deliberately? My only concern would be whether e2e tests rely on the date in some way for one of the test cases.

fake-ppud-automation-api/__files/get-offender-X098092.json
```diff
@@ -32,6 +32,7 @@
...
         "releases": [
```
ppmg, line 36: Shouldn't the `releases`be removed?

ppmg: And the same on line 75 of this file (GitHub won't let me comment on it because it's too far from changes you've made).

fake-delius-integration-api/__files/recommendation-model-X788206.json
ppmg, line 1: This is identical to some of the other json files in this same folder, is that right? If so, is there any point in having duplicates instead of having the mapping entries pointing to the same json and only keep separate json files if we have differences?

fake-ppud-automation-api/__files/search-PPCS_X098092.json
ppmg, line 1: Is there any point in having duplicates instead of having the mapping entries pointing to the same json and only keep separate json files if we have differences (i.e. remove this file and have the new `search-PPCS_X098092.json` file you're adding to the `mapping` folder point to `search-PPCS_X738925.json` instead)?

ppmg (changes requested):

fake-ppud-automation-api/mappings/search-PPCS_X098092.json
MancRickster: good spot, test removed

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/PpudOffender.kt
ppmg: Discussed: kept because existing recommendations with releases will fail on retrieval otherwise.

fake-ppud-automation-api/__files/get-offender-X098092.json
ppmg: Discussed: it did not affect the tests.

fake-delius-integration-api/__files/recommendation-model-X788206.json
ppmg: Discussed: keeping it this way for now, following the existing pattern. Will look into a clean-up when we're more familiar with the code.

ppmg (approved):
LGTM

PR #1117, opened 2024-12-13
Increases timeout to 90 minutes and reduces to once/day (6am mon-fri)…

andrewl:
… frequency of pg restore job


MancRickster (approved):
lgtm

PR #1120, opened 2024-12-19
MRD 2529 Support new OSP values

ppmg:
New sexual predictor scores, OSP/DC and OSP/IIC, were introduced earlier this year
by the ARN service to replace the old OSP/C and OSP/I ones. Older records will
continue to have the old scores, so we need to continue to support them while also
supporting the new ones.

If a record is provided with both old and new scores, the latter are prioritised
and the former ignored.


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/risk/converter/RiskScoreConverter.kt
```diff
@@ -0,0 +1,140 @@
...
+        riskScoreResponse.violencePredictorScore?.ovpRisk,
+        riskScoreResponse.violencePredictorScore?.oneYear,
+        riskScoreResponse.violencePredictorScore?.twoYears,
+        "OVP",
+      ),
+    )
+  }
+
+  private fun rsrLevelWithScore(riskScoreResponse: RiskScoreResponse?): LevelWithScore? {
+    val rsr = riskScoreResponse?.riskOfSeriousRecidivismScore
```
MancRickster, line 121: weird duplicate local ref of the same property

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/risk/converter/RiskScoreConverter.kt
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+
+  private fun createScores(riskScoreResponse: RiskScoreResponse): Scores? {
+    return Scores(
+      rsr = rsrLevelWithScore(riskScoreResponse),
+      ospc = buildLevelWithScore(
+        riskScoreResponse.sexualPredictorScore?.ospContactScoreLevel,
+        riskScoreResponse.sexualPredictorScore?.ospContactPercentageScore,
+        "OSP/C",
```
MancRickster: should these codes be constants somewhere?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/risk/converter/RiskScoreConverter.kt
```diff
@@ -0,0 +1,140 @@
...
+      )
+    }
+  }
+
+  private fun buildLevelWithScore(
+    level: String?, percentageScore: String?, type: String?,
+  ): LevelWithScore? {
+    val scoreIsNull = level == null && percentageScore == null
+    val notApplicableWithZeroPercentScorePresent =
+      level.equals(SCORE_NOT_APPLICABLE, ignoreCase = true) && percentageScore == "0"
```
MancRickster, line 141: Do these percentageScores come in from ARNs as strings? Would it be beneficial to cast them to numerical types once we have them?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/oasysarnapi/RiskScoreResponseTestBuilder.kt
```diff
@@ -103,5 +104,5 @@ fun violencePredictorScore(
 }
 
 private fun randomScoreLevel(): String {
-  return arrayOf("LOW", "MEDIUM", "HIGH").random()
+  return arrayOf("LOW", "MEDIUM", "HIGH", "VERY HIGH").random()
```
MancRickster, line 112: Should these be constants/enums somewhere?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/risk/converter/RiskScoreConverter.kt
ppmg: I wondered about this, but thought I'd discuss with you at some point to do in the future. Didn't want to make this a larger PR than it is already. I guess it's also debatable whether there is any benefit in converting it when all we're going to do is pass it through to the UI - either way, for now I don't think it does any harm leaving it as it was, so won't change it here.

MancRickster (approved):
LGTM

PR #1125, opened 2025-01-06
coping with facial image id lookup failures

MancRickster:
NOMIS can store invalid facial image IDs which will cause 404s during the offender retrieval.
This update protects against any failure in the non-essential image lookup


ppmg (approved):
LGTM

PR #1127, opened 2025-01-08
Suppression of CVE-2024-50379 trivy scan alert

MancRickster:
Suppression for tomcat vulnerability affecting jsp compilation in the default servlet can be suppressed as we do not use the default servlet and haven't configured it for write either


ppmg (approved):
LGTM

PR #1129, opened 2025-01-08
MRD-2609 Set 'Sentenced Under' field in PPUD Sentence

ppmg:
Support for setting a PPUD Sentence's 'Sentenced Under' field is added here,
both in storing it as part of a recommendation's bookRecallToPpud field as well
as in the endpoints redirecting PPUD actions to the PPUD Automation API
service.


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/ppud/PpudAutomationApiClient.kt
```diff
@@ -173,7 +173,7 @@ class PpudAutomationApiClient(
   private fun <T> post(url: String, request: Any, responseType: ParameterizedTypeReference<T>): Mono<T> {
     return webClient
       .post()
-      .uri { builder -> builder.path(url).build() }
+      .uri(url)
```
MancRickster, line 176: Wasn't the UrIBuilder used here in case `url` contains duplicate slashes, which `.path()` removes (or escapes depending on the default encoding)?

MancRickster (commented):
One tiny query

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/ppud/PpudAutomationApiClient.kt
ppmg: I'm not sure, to be honest. I noticed it when I was writing the unit test for this class - having it with the `UriBuilder` made things more complicated for the test and I couldn't see any reason why it was being done that way, so I switched it to the plainer String version.

I hadn't noticed the duplicate slashes behaviour. Maybe that was the reason when it was originally introduced, but I'm not sure it's necessary now. This is a private function and the other places it's called from in this file aren't doing any fancy path construction, they're simply inserting IDs into paths (i.e. we don't have things like `"user/$userId/$section"`, where section could have slashes). If we do encounter an ID with a slash, it might be wiser to error and investigate than assume we can just do away with the slash.

Let me know what you think. I can also be persuaded to switch it back (not sure what the unit test will look like in such a case, but can look into it) if you disagree with changing it.

MancRickster: Fair enough. 🙂

MancRickster (approved):
LGTM

PR #1131, opened 2025-01-10
MRD-2603 - additional field for CRN in PartA

MancRickster:
Additional field for CRN in section 4 of both preview and final PartA Word templates


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/DocumentData.kt
```diff
@@ -39,6 +39,7 @@ data class DocumentData(
   val ethnicity: String? = null,
   val croNumber: String? = null,
   val pncNumber: String? = null,
+  val crn: String? = null,
```
ppmg, line 42: Why nullable? Are there cases where we might not have a CRN?

MancRickster: Yer, it seems so - hence the firstname / surname search on the homepage.

ppmg (approved):
LGTM

PR #1133, opened 2025-01-14
Change default date time for nulls

ppmg:
The DEFAULT_DATE_TIME_FOR_NULL_VALUE is used whenever comparisons of entities by a date time is needed. Any entity with a null date time will be considered to have said constant as its date and time instead. The constant was set to epoch (1970-01-01T00:00:00) with the intention of null-date entities being considered older than those with non-null dates. This is being changed to an earlier date (in 1200), as there's still a (remote) possibility that an offender has an entity associated to them in the DB (e.g. an OSP assessment) earlier than epoch, which could result in incorrect ordering. This was noticed when a test using random values (RiskScoreConverterTest) produced early enough dates to result in unexpected ordering.

Note that using LocalDateTime.MIN.toString() was considered, but at the time of writing that returns a negative year. The '-' (minus) symbol can be represented with several different characters in ASCII, some of which have a greater lexicographic value than numbers do. Since the date time comparisons are done in strings, it was considered to be safer to use 1200 instead.


MancRickster (approved):
LGTM

PR #1135, opened 2025-01-16
MRD-2638 Stopping writing to the recommendations_history table 

MancRickster:
Stopping writing to the recommendations_history table and removal of all code referencing it (but not the DB table, nor column in recommendation_status)


ppmg:
The changes look fine to me. The only question I have is related to logging: the JIRA ticket says "we might want to look at increasing any logging we do in order to offset any loss of debugging information we currently use the history table for". Is this something that needs to be discussed? Looking at the RecommendationHistoryEntity, there are the `modified`, `modifiedBy` and `modifiedByUserFullName` fields, which would reflect data I don't think is necessarily reflected elsewhere (in particular if we have three people who work on a recommendation, only the creator and last one to update will be known in other places).

MancRickster:
Hi @ppmg , yer, I did think about this, but that would mean a new event for every update, which usually equates to a POST on every page, and we're already logging pageVisited events, so the only missing info is whether the user clicks Continue (POST) or Back, but even then, since we have the order of pageVisited events we can infer whether there was an update or not.

ppmg (approved):
LGTM

PR #1141, opened 2025-01-28
Mrd 2526 fixing info endpoint

MancRickster:
Reverting management endpoints (health/info/prometheus) back to standard HTTP ports.
These were overridden to use 8081 on 03/05/22 by Darren Oakley for MRD-11

### DIFF
--- a/docker-compose-e2e.yml
+++ b/docker-compose-e2e.yml
@@ -10,7 +10,7 @@ services:
       - "8080:8080"
       - "8081:8081"
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:8081/health/ping"]
+      test: ["CMD", "curl", "-f", "http://localhost:8080/health/ping"]
     environment:
       - SERVER_PORT=8080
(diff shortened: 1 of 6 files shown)


docker-compose-e2e.yml
```diff
@@ -10,7 +10,7 @@ services:
       - "8080:8080"
       - "8081:8081"
```
ppmg, line 11: Do we still need to expose 8081?

docker-compose.yml
```diff
@@ -10,7 +10,7 @@ services:
       - "8080:8080"
       - "8081:8081"
```
ppmg, line 11: Do we still need to expose 8081?

docker-compose-e2e.yml
ppmg: And I guess the same question goes for the Dockerfile file, which is exposing port 8081.

MancRickster: removed

docker-compose.yml
MancRickster: removed

ppmg (approved):
LGTM

PR #1144, opened 2025-01-29
MRD-2570 Ensuring we cope with inactive prisons (aka Progress Booking On when no Index Offence in NOMIS)

MancRickster:
Changing the endpoint used to retrieve prison/agency details by prisonId because the new endpoint facilitates querying inactive prisons (a possible scenario).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/PrisonApiClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/PrisonApiClient.kt
@@ -110,12 +110,12 @@ class PrisonApiClient(
     }
   }
 
-  fun retrieveAgency(agencyId: String, activeOnly: Boolean = false): Mono<Agency> {
+  fun retrieveAgency(agencyId: String): Mono<Agency> {
     val responseType = object : ParameterizedTypeReference<Agency>() {}
     return webClient
       .get()
       .uri { builder ->
-        builder.path("/api/agencies/$agencyId?activeOnly=$activeOnly").build()
+        builder.path("/api/agencies/$agencyId").queryParam("activeOnly", false).build()
       }
       .retrieve()
       .onStatus(
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/PrisonApiClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/PrisonApiClient.kt
@@ -110,12 +110,12 @@ class PrisonApiClient(
     }
   }
 
-  fun retrieveAgency(agencyId: String): Mono<Agency> {
+  fun retrieveAgency(agencyId: String, activeOnly: Boolean = false): Mono<Agency> {
     val responseType = object : ParameterizedTypeReference<Agency>() {}
     return webClient
       .get()
       .uri { builder ->
-        builder.path("/api/agencies/prison/" + agencyId).build()
+        builder.path("/api/agencies/$agencyId?activeOnly=$activeOnly").build()
       }
       .retrieve()
       .onStatus(
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/client/PrisonApiClientTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/client/PrisonApiClientTest.kt
@@ -114,7 +114,9 @@ class PrisonApiClientTest : IntegrationTestBase() {
 
   @Test
   fun `retrieve agency`() {
-    val request = HttpRequest.request().withPath("/api/agencies/MDI")
+    val request = HttpRequest.request()
+      .withPath("/api/agencies/MDI")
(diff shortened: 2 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/PrisonApiClient.kt
```diff
@@ -110,12 +110,12 @@ class PrisonApiClient(
     }
   }
 
-  fun retrieveAgency(agencyId: String): Mono<Agency> {
+  fun retrieveAgency(agencyId: String, activeOnly: Boolean = false): Mono<Agency> {
```
ppmg, line 113: Do we need to call it with `activeOnly = true` from anywhere? If not, maybe it's better to just keep it set to false within this function and not bother with extra parameters?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/PrisonApiClient.kt
```diff
@@ -110,12 +110,12 @@ class PrisonApiClient(
...
     val responseType = object : ParameterizedTypeReference<Agency>() {}
     return webClient
       .get()
       .uri { builder ->
-        builder.path("/api/agencies/prison/" + agencyId).build()
+        builder.path("/api/agencies/$agencyId?activeOnly=$activeOnly").build()
```
ppmg, line 118: I think you can do `path(...).queryParam("activeOnly", activeOnly)` so you don't need to worry about the syntax of adding a query parameter or if we add more params in the future, etc.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/PrisonApiClient.kt
```diff
@@ -110,12 +110,12 @@ class PrisonApiClient(
     }
   }
 
-  fun retrieveAgency(agencyId: String): Mono<Agency> {
+  fun retrieveAgency(agencyId: String, activeOnly: Boolean = false): Mono<Agency> {
     val responseType = object : ParameterizedTypeReference<Agency>() {}
```
ppmg, line 114: I think the new endpoint returns a `longDescription` instead of a `formatedDescription`, so the type definition will need to change the field name in order to capture the return value - unless I've missed some automagic going on somewhere?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/client/PrisonApiClientTest.kt
```diff
@@ -113,22 +113,36 @@ class PrisonApiClientTest : IntegrationTestBase() {
   }
 
   @Test
-  fun `retrieve agency`() {
-    val request = HttpRequest.request().withPath("/api/agencies/prison/MDI")
+  fun `retrieve agency, including inactive`() {
+    val request = HttpRequest.request().withPath("/api/agencies/MDI%3FactiveOnly=false")
```
ppmg, line 117: I think you can use `withQueryStringParameters` here if you want to avoid having to handle the query parameter syntax and encoding.

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/PrisonerApiServiceTest.kt
```diff
@@ -262,6 +263,66 @@ internal class PrisonerApiServiceTest : ServiceTestBase() {
...
+    given(prisonApiClient.retrieveOffender(any())).willReturn(
+      Mono.fromCallable {
+        offenderResponse
+      },
+    )
+
+    val result = PrisonerApiService(prisonApiClient).retrieveOffences(nomsId)
+
+    val offences = result.get(0).offences
+    assertThat(offences.size).isGreaterThan(0)
```
ppmg, line 323: Could you add some assertions here making sure we are still filling in the other fields and not just returning everything as null, please?

ppmg (changes requested):

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/client/PrisonApiClientTest.kt
```diff
@@ -113,29 +113,15 @@ class PrisonApiClientTest : IntegrationTestBase() {
...
-    val agency = prisonApiClient.retrieveAgency("MDI", false).block()
-
-    assertThat(agency?.agencyId, equalTo("MDI"))
-  }
-
-  @Test
-  fun `retrieve agency, active only`() {
-    val request = HttpRequest.request().withPath("/api/agencies/MDI%3FactiveOnly=true")
+  fun `retrieve agency`() {
+    val request = HttpRequest.request().withPath("/api/agencies/MDI")
```
ppmg, line 117: Small point: could you add `.withQueryStringParameter("activeOnly", false))` to this to ensure we are always requesting the inactive ones too, please?

ppmg (changes requested):

ppmg (approved):
LGTM

PR #1146, opened 2025-02-02
Mrd 2650 remove unused book recall endpoint

MancRickster:
Removing a deprecated endpoint whilst I was working in the area.
This project's 
`/ppud/book-recall/{nomisId `
call was being proxied on to PPUD API's
`/offender/$nomisId/recall`
but neither were called from the UI


MancRickster:
incorrect commits

PR #1147, opened 2025-02-03
Mrd 2650 remove unused book recall endpoint

MancRickster:
This project's
`/ppud/book-recall/{nomisId`
call was being proxied on to PPUD API's
`/offender//recall`
but neither were called from the UI


ppmg (changes requested):
I see the response classes have been removed, but what about the request one, `PpudBookRecall`? Is that being used elsewhere or can it also go?

ppmg (approved):
LGTM

PR #1153, opened 2025-02-28
Mrd 2634 redis caching lao calls

MancRickster:
Additional redis container on port 6369
Caching of the call for Limited Access Offender
Updating existing integration tests to cope with the new caching.
Additional integration test to test caching whether enabled or disabled


src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/CaseOverviewControllerTest.kt
```diff
@@ -23,6 +28,14 @@ class CaseOverviewControllerTest(
   @Value("\${ndelius.client.timeout}") private val nDeliusTimeout: Long,
 ) : IntegrationTestBase() {
 
+  @Autowired
+  lateinit var cacheManager: CacheManager
+
+  @BeforeEach
+  fun setup() {
+    cacheManager[USER_ACCESS_CACHE_KEY]?.invalidate()
```
ppmg, line 36: Why do we need to invalidate the cache before each test?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/OffenderSearchControllerTest.kt
```diff
@@ -91,6 +98,8 @@ class OffenderSearchControllerTest(
       val crn = "X123456"
       limitedAccessPractitionerOffenderSearchResponse(crn)
       userAccessExcluded(crn)
+      // Clear previous cache write
+      cacheManager[USER_ACCESS_CACHE_KEY]?.invalidate()
```
ppmg, line 102: Why do we need to clear the cache in this case?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/OffenderSearchControllerTest.kt
```diff
@@ -160,6 +169,8 @@ class OffenderSearchControllerTest(
       val crn = "X123456"
       limitedAccessPractitionerOffenderSearchResponse(crn)
       userNotFound(crn)
+      // Clear previous cache write
+      cacheManager[USER_ACCESS_CACHE_KEY]?.invalidate()
```
ppmg, line 173: Why do we need to clear the cache in this case?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/OffenderSearchControllerTest.kt
```diff
@@ -203,6 +214,8 @@ class OffenderSearchControllerTest(
       val crn = "X123456"
       limitedAccessPractitionerOffenderSearchResponse(crn)
       userAccessAllowed(crn, delaySeconds = nDeliusTimeout + 2)
+      // Clear previous cache write
+      cacheManager[USER_ACCESS_CACHE_KEY]?.invalidate()
```
ppmg, line 218: Why do we need to clear the cache in this case?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/PersonDetailsControllerTest.kt
```diff
@@ -79,6 +86,8 @@ class PersonDetailsControllerTest(
   fun `handles scenario where no person exists matching crn`() {
     runTest {
       userAccessAllowed(crn)
+      // Clear previous cache write
+      cacheManager[USER_ACCESS_CACHE_KEY]?.invalidate()
```
ppmg, line 90: Why do we need to clear the cache only in some cases in this test class?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/service/UserAccessValidatorTest.kt
```diff
@@ -0,0 +1,80 @@
...
+class UserAccessValidatorTest() : IntegrationTestBase() {
+
+  @Autowired
+  private lateinit var userAccessValidator: UserAccessValidator
+
+  @Autowired
+  lateinit var cacheManager: CacheManager
+
+  @Test
+  fun `checks user access - caching test`(@Value("\${spring.cache.type}") isCacheEnabled: String = "none") {
```
ppmg, line 29: I'm not sure what you're trying to get out of the `@Value` parameter here. As far as I understand (unless it works differently in tests), the `@Value` annotation will just find the value set on that property in the application configuration and inject it into the parameter; if the value isn't configured, it defaults it to `"none"`. So I think you will only get one of the two cases tested, whereas presumably we want both tested?

If we want both tested, I think it might be simpler to just have two separate tests within this test class and refactor the common part (the beginning of this test) so you don't have to repeat yourself in each tests.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/config/CacheConfiguration.kt
```diff
@@ -0,0 +1,51 @@
...
+
+  @Bean
+  fun cacheManager(
+    connectionFactory: RedisConnectionFactory,
+    @Value("\${spring.cache.type}") isCacheEnabled: String = "none",
+  ): CacheManager {
+    if (isCacheEnabled.equals("none", ignoreCase = true)) {
+      return NoOpCacheManager()
+    } else {
+      return RedisCacheManager.builder(connectionFactory)
```
ppmg, line 34: Sorry, I'm struggling to follow all the details of this configuration. I tried looking at the redis config in PPUD Automation, but that one seems to be simpler than this - why do they need to be different?

ppmg (changes requested):

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/OffenderSearchControllerTest.kt
MancRickster, line 218: Previous tests could have populated the cache, so we need the actual call (with delay) to be made to prove the timeout

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/OffenderSearchControllerTest.kt
MancRickster: Previous tests could have populated the cache, so we need the actual call (with 404 response) to be made to prove the NotFound

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/OffenderSearchControllerTest.kt
MancRickster: Previous tests could have populated the cache, so we need the actual call (with correct response) to be made to prove the test

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/CaseOverviewControllerTest.kt
MancRickster: All of the test were being affected by cache pre-population, so in order to preserve the initial aim of the test, I'm running them as if cache was disabled (or reliable cache misses)

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/service/UserAccessValidatorTest.kt
MancRickster: Yer, this was just testing the caching whether it was enabled or not (since AFAIK it's not possible to disable/re-route caching at runtime)

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/config/CacheConfiguration.kt
MancRickster: A couple of reasons: PPUD Aut only has one TTL for all cache keys, we will need to vary these; PPUD Aut caches strings whereas we need to cache objects (hence the additional serialisation config); PPUD Aut is always-on (I wanted to be able to set it via config using the standard Spring Boot method).

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/service/UserAccessValidatorTest.kt
ppmg: Right, so our tests all inherit from `IntegrationTestBase`, which has `@ActiveProfiles("test")` set. This means the tests run using the properties set in `application.yml`, with whichever ones appear in `application-test.yml` overriding the first (so it merges the two yml files, prioritising the latter).

What we can do is:
- Disable the cache in `application-test.yml`. This means we don't have to clear the cache in every test getting user access values.
- Split this test class into two classes, `UserAccessValidatorWithCacheEnabledTest` and `UserAccessValidatorWithCacheDisabledTest`. The latter can then be annotated with `@SpringBootTest(webEnvironment = RANDOM_PORT, properties = { "spring.cache.type=" })` (I think it's OK to set the property to nothing; I couldn't find examples online, though).

ppmg (changes requested):

.circleci/config.yml
```diff
@@ -45,15 +45,30 @@ slack-fail-post-step: &slack-fail-post-step
...
+        environment:
+          _JAVA_OPTIONS: -Xmx1024m -XX:ParallelGCThreads=2 -XX:ConcGCThreads=2
+            -XX:ParallelGCThreads=2
+            -Djava.util.concurrent.ForkJoinPool.common.parallelism=2
+            -Dorg.gradle.daemon=false
+      - image: postgres:14
+        environment:
+          - POSTGRES_PASSWORD=secret
+          - POSTGRES_USER=mrd_user
+          - POSTGRES_DB=make_recall_decision
```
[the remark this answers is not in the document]
MancRickster: 5432 is the default postgres port, so not needed

.circleci/config.yml
```diff
@@ -45,15 +45,30 @@ slack-fail-post-step: &slack-fail-post-step
             ]
           }
 
+executors:
+  java_postgres_redis:
+    docker:
+      - image: cimg/openjdk:21.0
+        environment:
+          _JAVA_OPTIONS: -Xmx1024m -XX:ParallelGCThreads=2 -XX:ConcGCThreads=2
+            -XX:ParallelGCThreads=2
```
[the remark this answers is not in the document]
MancRickster: Good spot, removed

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/config/CacheConfiguration.kt
```diff
@@ -0,0 +1,51 @@
...
+import java.time.Duration
+
+@EnableCaching
+@Configuration
+class CacheConfiguration(private val buildProperties: BuildProperties) {
+
+  @Bean
+  fun cacheManager(
+    connectionFactory: RedisConnectionFactory,
+    @Value("\${spring.cache.type}") isCacheEnabled: String = "none",
```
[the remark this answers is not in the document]
MancRickster: REDIS_CACHING_ENABLED env var removed.
Default corrected ('none' is the correct value)

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/config/CacheConfiguration.kt
MancRickster: Yes, we want TTL to be the same for all keys within the same cache, but we wont want all caches to share the same TTL, so these can be set in future with additional lines like:
`.withCacheConfiguration(USER_DETAILS_CACHE_KEY, getCacheConfiguration(Duration.ofHours(24)))`

MancRickster: In past projects it's been really useful to disable all caching in one place, rather than unpicking annotations from code

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/config/CacheConfiguration.kt
```diff
@@ -0,0 +1,51 @@
...
+    } else {
+      return RedisCacheManager.builder(connectionFactory)
+        .withCacheConfiguration(USER_ACCESS_CACHE_KEY, getCacheConfiguration(Duration.ofMinutes(60)))
+        .build()
+    }
+  }
+
+  private fun getCacheConfiguration(ttl: Duration): RedisCacheConfiguration {
+    val customObjectMapper = jacksonObjectMapper()
+      .activateDefaultTyping(jacksonObjectMapper().polymorphicTypeValidator, ObjectMapper.DefaultTyping.EVERYTHING, JsonTypeInfo.As.PROPERTY)
```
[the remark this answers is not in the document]
MancRickster: Yer, the suggested override `NON_FINAL_AND_ENUMS` causes the serialization to fail `Could not resolve subtype of [simple type, class java.lang.Object]: missing type id property '@class'`

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/config/CacheConfiguration.kt
```diff
@@ -0,0 +1,51 @@
...
+      .configure(FAIL_ON_UNKNOWN_PROPERTIES, false)
+    val jackson2JsonRedisSerializer = GenericJackson2JsonRedisSerializer(customObjectMapper)
+
+    return RedisCacheConfiguration.defaultCacheConfig()
+      .serializeValuesWith(RedisSerializationContext.SerializationPair.fromSerializer(jackson2JsonRedisSerializer))
+      .prefixCacheNameWith("${buildProperties.version}-")
```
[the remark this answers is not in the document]
MancRickster: To ensure cache recreation if/when the API build changes which could indicate cached objects fail deserialisation

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/service/UserAccessValidatorTest.kt
MancRickster: Good call. Update incoming...

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/config/CacheConfiguration.kt
```diff
@@ -26,7 +26,7 @@ class CacheConfiguration(private val buildProperties: BuildProperties) {
   @Bean
   fun cacheManager(
     connectionFactory: RedisConnectionFactory,
-    @Value("\${spring.cache.type}") isCacheEnabled: String = "none",
+    @Value("\${spring.cache.type:none}") isCacheEnabled: String,
   ): CacheManager {
     if (isCacheEnabled.equals("none", ignoreCase = true)) {
```
ppmg, line 31: Sorry, didn't think of this in the previous review: could you have the `else` case be an `else if` checking whether the cache type is set to `redis`, and have an `else` case that fails with something along the lines of `Unsupported cache type: <the-read-value>`, please? Just to make sure nobody simply changes `spring.cache.type` to some other caching library and thinks that's all they need to do to change the cache.

src/test/resources/application-test.yml
```diff
@@ -7,6 +7,9 @@ spring:
   flyway:
     enabled: false
 
+  cache:
```
ppmg, line 10: Could you add a comment here explaining we disable the cache because there are several tests relying on specific user access set-ups which would be ignored if there is data cached, please?

.circleci/config.yml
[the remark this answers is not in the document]
ppmg: We do refer to the port explicitly in various other places, so I think it would be safer to set it explicitly here. I know it's unlikely that postgres will change their default port, but at least if we set it here it's clear to future readers why it's set to that value in the other places too.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/config/CacheConfiguration.kt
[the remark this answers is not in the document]
ppmg: Ah, so this means that on recreation the existing cache is ignored and a new one built?

ppmg (approved):
LGTM

PR #1154, opened 2025-03-03
MRD-2639 Make ppudOffender.establishment optional

ppmg:
The establishment field in PpudOffender is made nullable/optional to prevent errors when loading older recommendations in the system whose ppudOffender field was retrieved before the establishment field was introduced. A separate ticket has been opened, MRD-2693, to eventually make the field non-nullable.


MancRickster (approved):
LGTM

PR #1156, opened 2025-03-11
MRD-2634 Reverting to standard redis port

MancRickster:
reverting to standard redis port since cache shared by all pods in same k8s env


ppmg (approved):
LGTM

PR #1158, opened 2025-03-12
Mrd 2634 - CircleCI redis config shared between projects

MancRickster:
Revert to stnd redis port
Explicitly configure the API to depend on the redis service setup in the UI project, and explicitly set the host via env. var.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -55,17 +55,6 @@ services:
       - POSTGRES_USER=mrd_user
       - POSTGRES_DB=make_recall_decision
 
-  redis:
-    image: redis:7.4.2-alpine3.21
-    ports:
-      - 6379:6379
-    expose:
-      - 6379
-    networks:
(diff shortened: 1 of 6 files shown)


docker-compose.yml
```diff
@@ -5,13 +5,16 @@ services:
       context: .
     networks:
       - hmpps
+    depends_on:
+      - redis
```
KiraNamida, line 9: If redis is no longer defined below, what is this resolving to?

MancRickster: The theory is that it'll reference the redis defined in the UI's docker-compose file which shares the same network. -if this doesn't work then I'll try setting the network to 'external:true' (doc isn't very clear on whether this is absolutely necessary)

KiraNamida: 👍

KiraNamida (approved):
:shipit:

PR #1161, opened 2025-03-12
MRD-2698 Support offender movement retrieval

ppmg:
A new endpoint is added for retrieving offender movements. These requests are
funneled through to the Prison API, where the data comes from, and refined
before returning (not all the data given by the Prison API is relevant to our
current use case).

These changes also include the introduction of ClientTimeoutRuntimeException,
and alternative to the existing ClientTimeoutException class. The latter has
been used so far when handling timeouts from calling other services (such as
the Prison API), but it extends Exception, not RuntimeException. The timeout
handling happens within the processing of a Mono through a callback register,
doOnError, which, in case of a new exception being raised, expects it to be a
RuntimeException. When this isn't the case, it wraps the raised exception with
its own ReactiveException.

So far, these wrapping ReactiveExceptions have been handled explicitly in our
code by the getValueAndHandleWrappedException method, which unwrap the timeout
exception and re-raise it. It is unclear to me why the timeout exception wasn't
made a RuntimeException (the original PR doesn't explain it). Since the new
runtime version works as expected, we'll look to replace the non-runtime one in
a separate set of changes (MRD-2706).

As part of these changes, some first steps towards refactoring out ServiceTestBase
have been carried out:
The ServiceTestBase class has become a place to put in any functionality that
may be re-used across more than one test class and is now bloated and unwieldy.
The inheritance pattern makes less sense for unit testing (as opposed to
integration testing), as each unit test should (ideally) be self-contained and
easily readable (which is harder if having to navigate across multiple classes
to understand it), so the ultimate goal is for the ServiceTestBase class to
disappear. We should try to split it out into multiple smaller auxiliary
classes for the parts it makes sense to keep and get rid of what is no longer
needed.

With these changes, we take the first step by removing PrisonerApiServiceTest's
inheritance of ServiceTestBase and using mockito for instantiation and
annotation-based injection of the mocks to avoid repeating the initialisation
of the class under test in each test case.


src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/prisonapi/domain/PrisonApiOffenderMovementTestBuilder.kt
```diff
@@ -0,0 +1,76 @@
...
+  movementDate,
+  movementTime,
+  movementReason,
+  movementReasonCode,
+  commentText,
+)
+
+internal fun PrisonApiOffenderMovement.toJson() = json(toJsonString())
+
+internal fun PrisonApiOffenderMovement.toJsonString() =
```
KiraNamida, line 56: Just for my own benefit, what is this used for that means we have need for this over the simple toJson() above? Is this used for a testing perhaps? Just seems a lot to maintain this manually and wondering what the history is.

Not a show stopper though as I am presume the answer will be "we always do this because it is used in..."

KiraNamida (dismissed):

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/prisonapi/domain/PrisonApiOffenderMovementTestBuilder.kt
ppmg: Not sure if I'm understanding fully, but I'll try to answer what I can so far: the `toJson` above is using this `toJsonString` method (i.e. there isn't already one which we're overriding or anything like that). I could just define a `toJson` method on its own, but it would still need some code that transforms the object into a JSON string.

As for the manual definitions, yes, you're right, it is a lot to maintain. I found a similar pattern being followed elsewhere in the code and have been using it ever since, but it certainly has its problems. One solution would be to [annotate the class to enable JSON serialization](https://kotlinlang.org/docs/serialization.html#serialize-and-deserialize-json), but I don't want to change production code for the sake of a test. However, prompted by your comment, I've started to look into the Gson library, which we might be able to use to simplify JSON string generation and make it more robust to changes. Wondering how Spring already does this (since we return this type through the Rest Controller, and it successfully turns it into a JSON), I've now noticed some `@JsonFormat`-annotated fields elsewhere in the code, so am going to look into this a bit further before pushing this through. Will update when I've learned more. Thanks for the question!

ppmg: OK, I've just updated it. I didn't use Gson in the end - we were already using a custom ObjectMapper, a class from the Jackson library for serializing and deserializing objects, so I've re-used that. I also changed a bit how we compare the two different offender movement classes to make it easier to read and more resilient to future changes. Hope it makes sense, but happy to clarify if needed!

KiraNamida (approved):
LGTM :shipit:

PR #1166, opened 2025-03-19
[MRD-2557]: PPCS e2e local journey

KiraNamida:
Changes to support MRD-2557, e2e local journey tests.

- Add new cypress env values to e2e_local_tests CircleCI job
- Some new wiremock files for PPCS journey
- A new spring profile to seed test data on setup. This is being used here in the local test to ensure we have a PPUD User mapping to retrieve else the attempt to send to PPUD returns a 404 response for failing to find this mapping even though it is only sending to a faked instance.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,7 +149,7 @@ jobs:
             set +e
 
             npx cypress run \
-              --env TAGS='not @E2E and not @smoke',USERNAME=${CYPRESS_USERNAME_local},PASSWORD=${CYPRESS_PASSWORD_local},USERNAME_SPO=${CYPRESS_USERNAME_SPO_local},PASSWORD_SPO=${CYPRESS_PASSWORD_SPO_local},USERNAME_ACO=${CYPRESS_USERNAME_ACO_local},PASSWORD_ACO=${CYPRESS_PASSWORD_ACO_local} \
+              --env TAGS='not @E2E and not @smoke',USERNAME=${CYPRESS_USERNAME_local},PASSWORD=${CYPRESS_PASSWORD_local},USERNAME_SPO=${CYPRESS_USERNAME_SPO_local},PASSWORD_SPO=${CYPRESS_PASSWORD_SPO_local},USERNAME_ACO=${CYPRESS_USERNAME_ACO_local},PASSWORD_ACO=${CYPRESS_PASSWORD_ACO_local},USERNAME_PPCS=${CYPRESS_USERNAME_PPCS_local},PASSWORD_PPCS=${CYPRESS_PASSWORD_PPCS_local},MAKE_RECALL_DECISION_API_URL=${CYPRESS_MAKE_RECALL_DECISION_API_URL_local},HMPPS_AUTH_EXTERNAL_URL=${CYPRESS_HMPPS_AUTH_EXTERNAL_URL_local} \
               --config-file e2e_tests/cypress.config.ts \
               --browser chrome \
               --record false \
--- a/fake-delius-integration-api/mappings/user-info-api.json
+++ b/fake-delius-integration-api/mappings/user-info-api.json
@@ -1,0 +1,14 @@
+{
+    "request": {
+      "method": "GET",
+      "urlPathPattern": "/user/make-recall-decision-api details"
+    },
+    "response": {
+      "status": 200,
+      "bodyFileName": "user-info-api.json",
+      "headers": {
+        "Content-Type": "application/json"
+      }
+    }
+  }
+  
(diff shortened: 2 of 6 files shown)


fake-delius-integration-api/mappings/user-info-api.json
```diff
@@ -0,0 +1,14 @@
+{
+    "request": {
+      "method": "GET",
+      "urlPathPattern": "/user/make-recall-decision-api details"
```
ppmg, line 4: Is this a valid pattern with the blank space? Or did you accidentally leave the `details` part in?

fake-ppud-automation-api/__files/search-PPCS_X738925.json
```diff
@@ -2,9 +2,9 @@
   "results": [    {
     "id": "4F6666656E64657269643D313731383138G725H664",
     "croNumber": "123456/12A",
-    "nomsId": "JG123POE",
-    "firstNames": "John",
-    "familyName": "Teal",
-    "dateOfBirth": "2024-10-16"
+    "nomsId": "XX123ABC",
```
ppmg, line 5: Why did you need to change these values? I don't think this is going to cause any issues, but that's because I think it's just displayed in one screen and never used anywhere else again, so just curious in case I'm missing something.

.circleci/config.yml
```diff
@@ -149,7 +149,7 @@ jobs:
             set +e
 
             npx cypress run \
-              --env TAGS='not @E2E and not @smoke',USERNAME=${CYPRESS_USERNAME_local},PASSWORD=${CYPRESS_PASSWORD_local},USERNAME_SPO=${CYPRESS_USERNAME_SPO_local},PASSWORD_SPO=${CYPRESS_PASSWORD_SPO_local},USERNAME_ACO=${CYPRESS_USERNAME_ACO_local},PASSWORD_ACO=${CYPRESS_PASSWORD_ACO_local} \
+              --env TAGS='not @E2E and not @smoke',USERNAME=${CYPRESS_USERNAME_local},PASSWORD=${CYPRESS_PASSWORD_local},USERNAME_SPO=${CYPRESS_USERNAME_SPO_local},PASSWORD_SPO=${CYPRESS_PASSWORD_SPO_local},USERNAME_ACO=${CYPRESS_USERNAME_ACO_local},PASSWORD_ACO=${CYPRESS_PASSWORD_ACO_local},USERNAME_PPCS=${CYPRESS_USERNAME_PPCS_local},PASSWORD_PPCS=${CYPRESS_PASSWORD_PPCS_local},MAKE_RECALL_DECISION_API_URL=${CYPRESS_MAKE_RECALL_DECISION_API_URL_local},HMPPS_AUTH_EXTERNAL_URL=${CYPRESS_HMPPS_AUTH_EXTERNAL_URL_local} \
```
ppmg, line 152: Just for my understanding: why do the MAKE_RECALL_DECISION_API_URL and HMPPS_AUTH_EXTERNAL_URL values need to be set now (i.e. why weren't they needed for the non-PPCS cases)?

ppmg (changes requested):

fake-delius-integration-api/mappings/user-info-api.json
KiraNamida: Cannot recall but you are correct that it definitely doesn't look right. Will investigate.

fake-ppud-automation-api/__files/search-PPCS_X738925.json
KiraNamida: This file was only being used for the PPCS journey but the data within didn't match the CRN that was being used to retrieve it. Honestly, I changed this because it was just confusing constantly seeing the results not being what was searched for.
Also, made for "and then get the result that matches the CRN being used" steps impossible as it just returned a different one.

.circleci/config.yml
KiraNamida: This is because this will now make direct calls to the api to "pre create" the recommendation in preparation for the PPCS journey test as a "Given a recommendation exists" test rather than "Given that the entire PO journey is fully functional before we can run this" test.
As such as use the service URL is to make create/update/update status calls directly to the UI. The hmpps auth url is used to retrieve a client access token to authorise these calls.

fake-delius-integration-api/mappings/user-info-api.json
KiraNamida: The more I look, the more I think this is a mistype that I just straight up missed amongst the everything else going on. Change coming in the next push.

ppmg (approved):
LGTM

PR #1169, opened 2025-04-04
MRD-2557: Update config for updated e2e test config

KiraNamida:
Brings e2e integrations tests in line with the new test config from the e2e solution


ppmg (approved):
LGTM

PR #1170, opened 2025-04-07
PI-2972 Migrate Delius search endpoints to integration service

marcus-bcl:
This change removes the dependency on probation-offender-search, and instead uses two new endpoints on the make-recall-decisions-and-delius integration API for finding a person on probation's details by name or CRN:
* POST /case-summary/search
* GET /case-summary/:crn

See https://github.com/ministryofjustice/hmpps-probation-integration-services/pull/4888


fake-delius-integration-api/__files/recommendation-model-X514364.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X514364 (in line with the file name and the mapping).

fake-delius-integration-api/__files/recommendation-model-X487027.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X487027 (in line with the file name and the mapping).

fake-delius-integration-api/__files/recommendation-model-X476202.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X476202 (in line with the file name and the mapping).

fake-delius-integration-api/__files/recommendation-model-X252642.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X252642 (in line with the file name and the mapping).

fake-delius-integration-api/__files/recommendation-model-X098099.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098099 (in line with the file name and the mapping).

fake-delius-integration-api/__files/recommendation-model-X098097.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098097 (in line with the file name and the mapping).

ppmg, line 9: This should be X098097 (in line with the file name and the mapping).

fake-delius-integration-api/__files/recommendation-model-X098096.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098096 (in line with the file name and the mapping).

fake-delius-integration-api/__files/recommendation-model-X098095.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098095 (in line with the file name and the mapping).

fake-delius-integration-api/__files/recommendation-model-X098092.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098092 (in line with the file name and the mapping).

fake-delius-integration-api/__files/recommendation-model-X098084.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098084 (in line with the file name and the mapping).

fake-delius-integration-api/__files/recommendation-model-X098083.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098083 (in line with the file name and the mapping).

fake-delius-integration-api/__files/recommendation-model-PPCS-X738925.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Dare"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X738925 (in line with the file name and the mapping).

fake-delius-integration-api/__files/personal-details_X788206.json
```diff
@@ -6,6 +6,7 @@
       "surname": "SmithX788206"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X788206 (in line with the file name and the mapping).

fake-delius-integration-api/__files/personal-details_PPCS_X738925.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Dare"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X738925 (in line with the file name and the mapping).

fake-delius-integration-api/__files/personal-details.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098092 (in line with the mapping).

fake-delius-integration-api/__files/personal-details-not-on-licence.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098083 (in line with the mapping).

fake-delius-integration-api/__files/personal-details-no-bespoke-licence-conditions.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098096 (in line with the mapping).

fake-delius-integration-api/__files/personal-details-no-additional-licence-conditions.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098097 (in line with the mapping).

fake-delius-integration-api/__files/licence-conditions_X788206.json
```diff
@@ -6,6 +6,7 @@
       "surname": "SmithX788206"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X788206 (in line with the file name and the mapping).

fake-delius-integration-api/__files/licence-conditions.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098092 (in line with the mapping).

fake-delius-integration-api/__files/licence-conditions-not-on-licence.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098083 (in line with the mapping).

fake-delius-integration-api/__files/licence-conditions-no-bespoke.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098096 (in line with the mapping).

fake-delius-integration-api/__files/licence-conditions-no-additional.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098097 (in line with the mapping).

fake-delius-integration-api/__files/licence-conditions-nd-and-cvl-same-start-dates.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098095 (in line with the mapping).

fake-delius-integration-api/__files/licence-conditions-more-recent-cvl.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098084 (in line with the mapping).

fake-delius-integration-api/__files/licence-conditions-more-recent-cvl-multiple-conditions.json
```diff
@@ -6,6 +6,7 @@
       "surname": "Smith"
     },
     "identifiers": {
+      "crn": "X000001",
```
ppmg, line 9: This should be X098099 (in line with the mapping).

docker-compose-postgres.yml
```diff
@@ -37,8 +37,6 @@ services:
       - 6379:6379
     expose:
       - 6379
-    networks:
```
ppmg, line 40: I already corrected this in a recent PR (by adding the network below, rather than removing this one here). You can leave this file untouched

ppmg (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/DeliusClient.kt
```diff
@@ -32,11 +35,20 @@ class DeliusClient(
...
   }
 
+  fun findByName(firstName: String?, surname: String?, page: Int, pageSize: Int) =
+    post<CasePage>(
+      "/case-summary/search",
+      body = FindByNameRequest(firstName, surname),
+      parameters = mapOf("page" to listOf(page), "size" to listOf(pageSize)),
+    ).body!!
+
+  fun findByCrn(crn: String) = get<PersonalDetailsOverview>("/case-summary/$crn") { Mono.empty() }?.body
```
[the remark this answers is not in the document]
marcus-bcl: Yes, it will map to an empty Mono on 404 - so will be returned as null.  This makes it easier to handle a CRN not being found in the OffenderSearchService.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/DeliusClient.kt
```diff
@@ -67,20 +79,19 @@ class DeliusClient(
...
+    crossinline notFoundExceptionFunction: (String) -> Mono<Throwable> = { Mono.error(PersonNotFoundException(it)) },
+  ) = get<T>(endpoint, parameters, notFoundExceptionFunction)?.body!!
 
   private inline fun <reified T : Any> get(
     endpoint: String,
     parameters: Map<String, List<Any>> = emptyMap(),
-    crossinline notFoundExceptionFunction: (String) -> Throwable,
-  ): ResponseEntity<T> {
+    crossinline notFoundExceptionFunction: (String) -> Mono<Throwable>,
+  ): ResponseEntity<T>? {
```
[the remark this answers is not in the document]
marcus-bcl: Yes exactly, it's to allow 404 to be treated as null.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/DeliusClient.kt
```diff
@@ -143,13 +175,24 @@ class DeliusClient(
...
     data class Identifiers(
+      val crn: String,
       val pncNumber: String?,
       val croNumber: String?,
       val nomsNumber: String?,
       val bookingNumber: String?,
     )
   }
 
+  data class FindByNameRequest(
```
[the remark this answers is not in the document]
marcus-bcl: Good spot. I'll update it to make forename and surname non-nullable. As far as I can tell the UI validates that both are provided anyway.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/OffenderSearchService.kt
```diff
@@ -21,65 +18,36 @@ internal class OffenderSearchService(
...
-      AccessRestrictions(
-        userExcluded = userAccessResponse.userExcluded,
-        userRestricted = userAccessResponse.userRestricted,
-      )
-    } else {
-      AccessRestrictions(userExcluded = false, userRestricted = false)
-    }
-  }
+  private fun DeliusClient.PersonalDetailsOverview.toOffenderSearchOffender(): OffenderSearchOffender {
+    val access = userAccessValidator.checkUserAccess(identifiers.crn)
```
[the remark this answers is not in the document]
marcus-bcl: As I understand it, the previous code assumed that a null or blank name from the search API indicated that there were some access restrictions on the case, but without checking whether the specific user could access it - hence the follow-up call to check against the specific user. So the search API did _some_ redaction, but didn't take into account the user.

For the probation integration APIs, we typically don't do any redaction or blanking-out of fields ourselves, and instead provide a separate endpoint for checking user access - to allow clients to perform appropriate redaction in their service.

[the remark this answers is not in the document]
marcus-bcl: Are you saying you don't need the user access check at all?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/PpcsServiceTest.kt
```diff
@@ -25,59 +22,67 @@ internal class PpcsServiceTest : ServiceTestBase() {
...
+        primaryLanguage = "test",
+      ),
     )
+    given(deliusClient.getUserAccess(username, "X90902")).willReturn(noAccessLimitations())
 
-    val result = PpcsService(recommendationRepository, recommendationStatusRepository, offenderSearchApiClient)
+    val result = PpcsService(recommendationRepository, recommendationStatusRepository, deliusClient, userAccessValidator)
+      .search("X90902")
+
+    assertThat(result.results).isEmpty()
```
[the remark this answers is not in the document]
marcus-bcl: Good point, it's being dropped from the results because there is no recommendation doc. But the test should be checking it's dropped due to an exclusion. I'll update this.

fake-delius-integration-api/mappings/search-multiple-results-page1.json
```diff
@@ -1,7 +1,7 @@
 {
   "request": {
     "method": "POST",
-    "url": "/search/people?page=1&size=20",
+    "url": "/case-summary/search?page=1&size=20",
     "bodyPatterns": [
       {
         "matchesJsonPath": "[?($.surname =~ /o'connor/i)]"
```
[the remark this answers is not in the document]
marcus-bcl: Good spot, thanks!

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/OffenderSearchService.kt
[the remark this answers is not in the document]
ppmg: We probably do, but I'll need to double check with my BA. I first wanted to check whether there's a change in behaviour happening here or not, as I don't know whether the integration API behaves differently from the Delius one. Does the Delius API return blank names if the user is excluded or restricted from seeing that offender's information? Or is that an incorrect assumption our code is making (i.e. are we sometimes returning offender data a particular user shouldn't see because Delius didn't blank the names out and our code assumes that means it's fine to show?)?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/DeliusClient.kt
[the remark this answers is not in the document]
ppmg: Sorry, I think I've gone a bit down a rabbit hole on this and maybe I'm missing the forest for the trees: do the new API endpoints behave in the same way as the old ones when no results are found for a given set of search values (either name or CRN) and, if so, what is that behaviour? Return 200 with an empty body (or empty page) or return 404?

I was trying to find what the current behaviour is in our service and saw that we don't have tests for when the offender search API returns 404, so I wrote up ones to make it clear what the current behaviour is: https://github.com/ministryofjustice/make-recall-decision-api/commit/de9554a434883a4d4430754a5b287d5832ba73bb . My concern is whether we're preserving the behaviour or not (in case we also need to change the UI code, which consumes the CaR search endpoints).

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/OffenderSearchService.kt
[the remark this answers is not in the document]
marcus-bcl: I've had a look at the Probation Offender Search API code, and it wouldn't have performed any access checks on the endpoint you were calling. So the name would never be returned as blank. I've checked this in dev for a limited access case as well, and can confirm it comes back with a populated name - so I think the assumption was wrong.

It would be good to get this change in and test that it now behaves correctly for limited access cases, however I'm also happy to remove it to preserve the previous behaviour if you prefer?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/PpcsServiceTest.kt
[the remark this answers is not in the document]
marcus-bcl: Can do, although I would have to update the Mockito settings to be lenient - because it never actually gets to the recommendation doc check.

I've added a verification at the end to assert that it never calls the recommendationRepository, which I think is cleaner - but let me know if you'd prefer to stick with your approach of mocking the call.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/DeliusClient.kt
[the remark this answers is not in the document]
marcus-bcl: The Probation Offender Search API would never have returned a 404, just an empty list. The new endpoint returns a 404 for the CRN lookup, which makes more sense because the CRN is a unique identifier. You can only have 0 or 1 result. So now there is either a single result returned (200), or no result (404).

The code change here is to handle that 404 and treat it as null, so the rest of the code can behave the same as it did before.

PR #1180, opened 2025-05-29
Handle Delius Search by CRN not found responses

KiraNamida:
Correct for differing not found response from the delius' `case-sumamry/{crn}` end point.

Following the update to the delius endpoint for searching by CRN, we know get two differently shaped results based on whether a user is found or not.

Not found: ```{ "status": 404, "message": "..." } ```

Found: ```{ "name": { ...} , ...etc }```

In order to contain the change, I have expanded the client to return a `SearchByCRNResponse` which combines both of these onto a single object and then we either detect the 404 and return null (which we previously received for a Not Found result) or reliably cast to the previous PersonalSummaryOverview and continue as we did before.


KiraNamida:
Thinking we might need to handle this a different way as it appears the client interrupts a 404 and throws a WebClientException.NotFound.
Re-engineering tests to mimic this and then will handle.

PR #1181, opened 2025-05-30
[MRD-2748] Handle search by CRN 404 response

KiraNamida:
[MRD-2748]()http://dsdmoj.atlassian.net/browse/MRD-2748)

Delius /case-summary/{crn} now returns a 404 rather than an empty response when not found, handle this and return the already existing null when this occurs.

[MRD-2748]: https://dsdmoj.atlassian.net/browse/MRD-2748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/fake-delius-integration-api/mappings/search-not-found.json
+++ b/fake-delius-integration-api/mappings/search-not-found.json
@@ -4,7 +4,7 @@
     "urlPath": "/case-summary/X123456"
   },
   "response": {
-    "status": 400,
+    "status": 404,
     "bodyFileName": "search-not-found.json",
     "headers": {
       "Content-Type": "application/json"
--- a/fake-delius-integration-api/mappings/search-not-found.json
+++ b/fake-delius-integration-api/mappings/search-not-found.json
@@ -1,0 +1,13 @@
+{
+  "request": {
+    "method": "GET",
+    "urlPath": "/case-summary/X123456"
+  },
+  "response": {
+    "status": 400,
+    "bodyFileName": "search-not-found.json",
+    "headers": {
+      "Content-Type": "application/json"
+    }
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/DeliusClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/DeliusClient.kt
@@ -3,6 +3,7 @@ package uk.gov.justice.digital.hmpps.makerecalldecisionapi.client
 import com.fasterxml.jackson.annotation.JsonFormat
 import io.micrometer.core.instrument.Counter
 import org.apache.commons.lang3.StringUtils.normalizeSpace
+import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.core.io.Resource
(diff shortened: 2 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/DeliusClient.kt
```diff
@@ -41,7 +42,14 @@ class DeliusClient(
     parameters = mapOf("page" to listOf(page), "size" to listOf(pageSize)),
   ).body!!
 
-  fun findByCrn(crn: String) = get<PersonalDetailsOverview>("/case-summary/$crn") { Mono.empty() }?.body
+  fun findByCrn(crn: String): PersonalDetailsOverview? {
+    val response = try {
+      get<PersonalDetailsOverview>("/case-summary/$crn") { Mono.empty() }?.body
```
ppmg, line 47: The lambda being passed in meant to deal with the NOT_FOUND status case. Do you know why it isn't working in this case? Or am I misunderstanding the problem?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/client/DeliusClientTest.kt
```diff
@@ -25,7 +25,8 @@ class DeliusClientTest : IntegrationTestBase() {
 
   @Test
   fun `find by crn returns null when not found`() {
-    val response = deliusClient.findByCrn("X123456")
+    findByCrnNotFound(crn = "X654321")
```
ppmg, line 28: My understanding is that wiremock should be returning 404 by default if it receives an unmapped request. Was this not the case? I'm OK with still adding this code (either that or adding a comment explicitly stating the reason why we don't make any wiremock config calls), but I'm wondering if it isn't doing so (and why).

fake-delius-integration-api/mappings/search-not-found.json
```diff
@@ -0,0 +1,13 @@
+{
+  "request": {
+    "method": "GET",
+    "urlPath": "/case-summary/X123456"
+  },
+  "response": {
+    "status": 400,
```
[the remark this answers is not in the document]
KiraNamida: Correct, I think I changed this to 400 to test it *wasn't* getting caught and didn't change it back. Thanks for the catch.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/DeliusClient.kt
KiraNamida: As best I could determine, between the behaviour being shown and rummaging in this client source code:
This is because it trying to first cast the body to a PersonalDetailsOverview which fails because the body we get with a 404 has the fields `status` and `message` so it's never getting to the handler as an exception is being thrown.

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/client/DeliusClientTest.kt
KiraNamida: This is to be explicit for this test case as it states the desired behaviour rather than relying on a default. I mostly did this for my own confidence when exploring how the client behaved and thought "might as well leave that there as it self documents".

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/DeliusClient.kt
KiraNamida: For additional reference: I found this being handled in a similar way in the Risk Service (https://github.com/ministryofjustice/make-recall-decision-api/blob/ea7f48c4ff1161acf1e435a653b77ce1c2a32b3e/src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/risk/RiskService.kt#L167) from which I took cues.

ppmg: I wasn't expecting it to try to cast in that case, but maybe I've missed how some of this works. Let's get this fix in and can look into the problem further if needed.

PR #1194, opened 2025-07-14
Add support for editing indeterminate sentence data

KiraNamida:
Add a new optional field to BookRecallToPpud. This `IndeterminateSentenceData` holds the working values of the sentence selected to allow user edits to the data before submission.


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/BookRecallToPpud.kt
```diff
@@ -14,6 +14,7 @@ data class BookRecallToPpud(
   val indexOffence: String? = null,
   val indexOffenceComment: String? = null,
   val ppudSentenceId: String? = null,
+  val ppudSentenceData: IndeterminateSentenceData? = null,
```
ppmg, line 17: I think the field name should also include the word 'indeterminate', to reduce the chances of taking from it for determinate cases (and maybe we eventually define a similar field for determinate cases?).

ppmg (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/BookRecallToPpud.kt
KiraNamida: I see what you mean but does that disassociate it from the ppudSentenceId which this data relates to? I wander if there is something between the two? 🤔

ppmg (approved):
LGTM. Pending point on naming of new field can be addressed in follow-up if needed, given it's an optional field on a non-live part of the code.

PR #1195, opened 2025-07-14
MRD-2684 Allow indeterminate sentences to be booked

ppmg:
PPUD sentence-related types are updated here with a tariffExpiryDate field in
line with updates made to PPUD Automation.

Mocking of PPUD Automation server responses has been refactored out of
IntegrationTestBase and into its own class. This is one of several steps needed
to move away from IntegrationTestBase being a god class.


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/PpudCreateOrUpdateSentenceRequest.kt
```diff
@@ -6,12 +6,12 @@ data class PpudCreateOrUpdateSentenceRequest(
...
   val licenceExpiryDate: LocalDate?,
-  val mappaLevel: String,
+  val mappaLevel: String?,
   val releaseDate: LocalDate?,
   val sentenceLength: SentenceLength?,
-  val espCustodialPeriod: PpudYearMonth? = null,
-  val espExtendedPeriod: PpudYearMonth? = null,
-  val sentenceExpiryDate: LocalDate? = null,
+  val espCustodialPeriod: PpudYearMonth? = null, // never set in UI - should we remove it and all associated code?
+  val espExtendedPeriod: PpudYearMonth? = null, // never set in UI - should we remove it and all associated code?
```
KiraNamida, line 13: Can I presume there are other fields that exist within PPUD that we do no concern ourselves with setting like these two.
I am wondering if this object represents the complete scope of creating a sentence or if it is already the partial set that pertains to CaR. If it's the later, then I would definitely move to remove these are that would make them demonstrably obsolete.

KiraNamida (approved):
Approving as outstanding conversation is just some existing clean up and does not effect this functionality.

fincardy-moj (approved):

PR #1197, opened 2025-07-18
Update indeterminate field names on BookRecallToPpud to better reflec…

KiraNamida:
…t the data they are for


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/BookRecallToPpud.kt
```diff
@@ -13,8 +13,8 @@ data class BookRecallToPpud(
   val releasingPrison: String? = null,
   val indexOffence: String? = null,
   val indexOffenceComment: String? = null,
-  val ppudSentenceId: String? = null,
-  val ppudSentenceData: IndeterminateSentenceData? = null,
+  val ppudIndeterminateSentenceId: String? = null,
```
ppmg, line 16: We need to also keep `ppudSentenceId`, as it is used for Determinate sentences (I re-used it without thinking too much about how we might be representing the rest of the indeterminate-specific data, sorry).

ppmg (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/BookRecallToPpud.kt
KiraNamida: I would say that is worth doing but is going to have significant functional changes that I am unlikely to get done before the end of today.
Worth considering a but of a restructure around this as overlapping data can lead to stale data if not managed correctly where as two distinct fields should better protect against that chance.
Shall we come up with a ticket to capture this?

PR #1203, opened 2025-08-04
MRD-2774: Update CaR to reflect updated FTR suitability and exclusion…

EmilyHazlehurst:
… criteria

- Add new fields to recommendation API and to tests


KiraNamida (changes requested):
It looks like there's a test in TemplateReplacementServiceTest to count the number of entries in the template map that needs updating/the new entires being checked.

EmilyHazlehurst:
> It looks like there's a test in TemplateReplacementServiceTest to count the number of entries in the template map that needs updating/the new entires being checked.

Thanks @KiraNamida, I've updated the test and it's passing now.

PR #1204, opened 2025-08-07
MRD-2776: Update FTR48 Part Template for new exclusion criteria

KiraNamida:
[MRD-2776](https://dsdmoj.atlassian.net/browse/MRD-2776): Update FTR48 Part Template for new exclusion criteria

Screenshot of updated template file `NAT Recall Part A London Template - obtained 231114 - v2025-07-29.docx`:
Updated Part A Template

Exclusion Criteria used for demonstrations:
Exclusion Criteria

Report generated using the above criteria and the new template:
Report created with updated template


EmilyHazlehurst:
Thanks @KiraNamida. A couple of things:

1. Could you update the preview file as well? _Preview NAT Recall Part A London Template - obtained 231114 - v2025-07-29.docx_
2. For [MRD-2761](https://dsdmoj.atlassian.net/browse/MRD-2761) Pablo changed the way we retrieve templates so that it's based on the latest date. 
Could you keep the old version of the file, make a new one, then change the version to today's date, i.e. _NAT Recall Part A London Template - obtained 231114 - **v2025-07-29**.docx_
Then add it to the `document-template` config in [src/main/resources/application.yml](https://github.com/ministryofjustice/make-recall-decision-api/pull/1200/files#diff-7bc26be01d3e7c9c566f2a28dc1b5cd87dbdd5cb619184f46eb2bf223bf7825d)

If it's at all possible to use a different template depending on the feature flag then that would be great, but if that's too complicated with the current set up then that can be covered in https://dsdmoj.atlassian.net/browse/MRD-2809

In which case we just have to be conscious not to release anything until we've made those improvements.

KiraNamida:
Okay I think I see what is going on here, I'll shuffle those updates around to slot into this system.

PR #1210, opened 2025-08-21
MRD-2807: Reorganise versioned templates ahead of FTR48 release

KiraNamida:
Reorganise document templating versioning and retrieve based on PP_DOCUMENT_CREATED if status exists for recommendation

----------
Previously, we put together a system where updated documents where added to resources and we had some config to select the named file based on the current time and the most recent start date in the past. Now we are making a few enhancements in preparation for continuing versioning to support future document updates.

- Separate files by version
  Each version of a file is being moved to it's own directory* named** for the version*** and given a predetermined name. This means that each document type can be versioned individually.
  The "new version" for Part A are copies of the files `NAT Recall Part A London Template - obtained 231114 - v2025-08-07.docx` and `Preview NAT Recall Part A London Template - obtained 231114 - v2025-08-07.docx` which have previously been prepared for the FTR48 release on Sept 2<sup>nd</sup>.
  \* Part A and Preview Part A documents are considered "the same" for this determination as their content is kept the same and an update to one requires an update to the other.
  \** The first instance references the intended release date and brief description for context as `2025-09-02 - FTR48 Phase 1`
  \*** This name is also the key set in config that will be used to look up the template based on the date resolution.
- Add a default version
  If no config is set for a template type (or the current setting is in the future and so doesn't resolve), default will be relied upon. These are files otherwise known as `NAT Recall Part A London Template - obtained 231114 - v2025-07-29.docx`, `Preview NAT Recall Part A London Template - obtained 231114 - v2025-07-29.docx` and `DNTR Template.docx` i.e. the documents already being used in Production so a deployment will maintain these as the defaults until the first version period begins and we begin respecting version dates.
- Template names
  The actual templates are now all named `Part A Template.docx`, ` Preview Part A Template.docx` and `DNTR Template.docx` so that the file plucked for each version is standardised. What the template is a part of has been retained in the names even through this is also a part of the parent directories so that the files do not immediately lose context if viewed from anywhere other than file tree.
- Prep config for prod release
  Now that we rely on default until a version is configured, version config has been reduced to the application yaml expecting the updated Part A/Preview Part A to go live for 2025-09-02 and dev overriding this to yesterday so we can dev test ahead of the release.
  We no longer need the "default entry [0]" we previously added for prod as a backstop as the default now exists en lieu of specific config.


src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/documenttemplate/TemplateRetrievalServiceTest.kt
```diff
@@ -20,60 +25,310 @@ class TemplateRetrievalServiceTest {
...
+
+  val noExistingDocument: ZonedDateTime? = null
+
+  @Nested
+  @DisplayName("Part A Documents")
+  inner class PartADocument {
+    @Test
+    fun `retrieves the default Part A template when no settings provided`() {
+      testLoadDocumentTemplate(
+        noDocumentsTemplateConfiguration,
```
ppmg, line 64: Is it worth testing the cases with no config set but with both past and future existing dates, to check it still goes for the default?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/documenttemplate/TemplateRetrievalServiceTest.kt
```diff
@@ -20,60 +25,310 @@ class TemplateRetrievalServiceTest {
...
+        DocumentType.PART_A_DOCUMENT,
+        noExistingDocument,
+        defaultDocumentTemplateSetting,
+      )
+    }
+
+    @Test
+    fun `retrieves the default Part A template when only configured setting is in the future`() {
+      testLoadDocumentTemplate(
+        futureOnlyDocumentsTemplateConfiguration,
```
ppmg, line 74: Is it worth having a test with future-only configuration and both future and no existing dates, to check it goes for the default?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/documenttemplate/TemplateRetrievalServiceTest.kt
```diff
@@ -20,60 +25,310 @@ class TemplateRetrievalServiceTest {
...
+  )
+  private val futureOnlyDocumentsTemplateConfiguration = documentTemplateConfiguration(
+    partATemplateSettings = listOf(futureDocumentTemplateSetting),
+    partAPreviewTemplateSettings = listOf(futureDocumentTemplateSetting),
+    dntrTemplateSettings = listOf(futureDocumentTemplateSetting),
+  )
+
+  val noExistingDocument: ZonedDateTime? = null
+
+  @Nested
```
ppmg, line 58: You can reduce duplication using
    @ParameterizedTest
    @EnumSource(DocumentType::class.java)

(unless you find the nesting version clearer)

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/config/documenttemplate/DocumentTemplateConfiguration.kt
```diff
@@ -12,13 +12,14 @@ import java.time.ZonedDateTime
  */
 @ConfigurationProperties(prefix = "document-template")
 data class DocumentTemplateConfiguration(
-  val partATemplateSettings: List<DocumentTemplateSetting>,
-  val partAPreviewTemplateSettings: List<DocumentTemplateSetting>,
-  val dntrTemplateSettings: List<DocumentTemplateSetting>,
+  val partATemplateSettings: List<DocumentTemplateSetting> = listOf(),
+  val partAPreviewTemplateSettings: List<DocumentTemplateSetting> = listOf(),
```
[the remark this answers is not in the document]
KiraNamida: That would depend if we are truly comfortable committing to them being in the same folder and in lock step. Though even as I type this, the more I think that isn't a terrible idea given the changes we're putting in for versioning as a whole so I might have convinced myself in doing so. 🤔

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/documenttemplate/TemplateRetrievalServiceTest.kt
```diff
@@ -20,60 +25,310 @@ class TemplateRetrievalServiceTest {
...
 
-    val expectedTemplateName = currentDocumentTemplateSetting.templateName
-    val expectedClassPathResource = ClassPathResource(expectedTemplateName)
+    val expectedTemplatePath = when (documentType) {
+      DocumentType.PART_A_DOCUMENT -> "partA/${expectedSettings.templateName}/Part A Template.docx"
+      DocumentType.PREVIEW_PART_A_DOCUMENT -> "partA/${expectedSettings.templateName}/Preview Part A Template.docx"
+      DocumentType.DNTR_DOCUMENT -> "dntr/${expectedSettings.templateName}/DNTR Template.docx"
+    }
+    val expectedClassPathResource = ClassPathResource("templates/$expectedTemplatePath")
+    val createdDateLocal = documentCreated?.toLocalDateTime()?.plusHours(1)
```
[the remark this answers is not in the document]
KiraNamida: The template config date must be _before_ the document created data and seeing as we're using the same past/present/future dates this was resulting in the config and created date being _equal_ so resolving the config once time frame back.
This was to avoid having to repeatedly tack this on the created dates above in all tests but this might be less of a concern if I can move to parametrised tests as I can probably centralise where this happens before it comes in to this method as the created date.

src/main/resources/application.yml
```diff
@@ -229,22 +229,14 @@ spring:
...
-
-  dntr-template-settings[0].start-date-time: "2025-01-31T14:00Z"
-  dntr-template-settings[0].template-name: "DNTR Template.docx"
+  # Any setting not specified or occurring only in the future will default to the 'default' template name
+  # part-a-template-settings[x].start-date-time/template-name # Settings for Part A templates
+  # part-a-preview-template-settings[x].start-date-time/template-name # Settings for Part A Preview templates
+  # dntr-template-settings[x].start-date-time/template-name # Settings for DNTR templates
+  part-a-template-settings[0].start-date-time: "2025-09-01T23:00Z" #2025-09-02T00:00Z in BST
+  part-a-template-settings[0].template-name: "2025-09-02 - FTR48 Phase 1"
+  part-a-preview-template-settings[0].start-date-time: "2025-09-01T23:00Z" #2025-09-02T00:00Z in BST
```
ppmg, line 238: The preview config values are now obsolete, aren't they?

helm_deploy/values-dev.yaml
```diff
@@ -21,18 +21,10 @@ generic-service:
...
-    DOCUMENTTEMPLATE_PARTAPREVIEWTEMPLATESETTINGS_0_STARTDATETIME: "2025-01-10T15:58Z"
-    DOCUMENTTEMPLATE_PARTAPREVIEWTEMPLATESETTINGS_0_TEMPLATENAME: "Preview NAT Recall Part A London Template - obtained 231114.docx"
-    DOCUMENTTEMPLATE_PARTAPREVIEWTEMPLATESETTINGS_1_STARTDATETIME: "2025-07-30T12:45Z"
-    DOCUMENTTEMPLATE_PARTAPREVIEWTEMPLATESETTINGS_1_TEMPLATENAME: "Preview NAT Recall Part A London Template - obtained 231114 - v2025-07-29.docx"
-    DOCUMENTTEMPLATE_PARTAPREVIEWTEMPLATESETTINGS_2_STARTDATETIME: "2025-08-07T23:00Z"
-    DOCUMENTTEMPLATE_PARTAPREVIEWTEMPLATESETTINGS_2_TEMPLATENAME: "Preview NAT Recall Part A London Template - obtained 231114 - v2025-08-07.docx"
+    DOCUMENTTEMPLATE_PARTATEMPLATESETTINGS_0_STARTDATETIME: "2025-08-20T23:00Z" # 2025-08-21T00:00Z in BST, bring forward dev for testing
+    DOCUMENTTEMPLATE_PARTATEMPLATESETTINGS_0_TEMPLATENAME: "2025-09-02 - FTR48 Phase 1"
+    DOCUMENTTEMPLATE_PARTAPREVIEWTEMPLATESETTINGS_0_STARTDATETIME: "2025-08-20T23:00Z" # 2025-08-21T00:00Z in BST, bring forward dev for testing
```
ppmg, line 26: The preview config values are now obsolete, aren't they?

helm_deploy/values-preprod.yaml
```diff
@@ -18,6 +18,11 @@ generic-service:
     MRD_URL: https://make-recall-decision-preprod.hmpps.service.justice.gov.uk
     MRD_API_URL: https://make-recall-decision-api-preprod.hmpps.service.justice.gov.uk
     HOUSEKEEPING_SENDDOMAINEVENTS: true
+    # we cannot override a single item in a list - the whole list is replaced, so we define the whole thing again
```
ppmg, line 21: These overrides use the old structure and document names and need updating (and the preview overrides removed, I think?).

helm_deploy/values-prod.yaml
```diff
@@ -17,6 +17,11 @@ generic-service:
     MRD_URL: https://make-recall-decision.hmpps.service.justice.gov.uk
     MRD_API_URL: https://make-recall-decision-api.hmpps.service.justice.gov.uk
     HOUSEKEEPING_SENDDOMAINEVENTS: true
+    # we cannot override a single item in a list - the whole list is replaced, so we define the whole thing again
```
ppmg, line 20: Do we want to override these config values at all in prod now?

src/main/resources/application.yml
KiraNamida: I think this is the merge conflict no resolving the way I thought it did. I am going to pull latest and manually check this. Same for the rest of the comments below on this block.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/util/DateTimeHelper.kt
```diff
@@ -26,14 +26,16 @@ class DateTimeHelper {
...
+
+    fun dateTimeUTCZonedFromString(utcDateTimeString: String?): ZonedDateTime = if (utcDateTimeString.isNullOrBlank()) {
+      ZonedDateTime.of(LocalDateTime.ofEpochSecond(0, 0, ZoneOffset.UTC), ZoneId.of("UTC"))
     } else {
       val utcZonedDateTime = ZonedDateTime.of(
         LocalDateTime.parse(utcDateTimeString.replace("Z", MrdTextConstants.EMPTY_STRING)),
         ZoneId.of("UTC"),
       )
-      utcZonedDateTime.withZoneSameInstant(ZoneId.of("Europe/London")).toLocalDateTime()
+      utcZonedDateTime.withZoneSameInstant(ZoneId.of("Europe/London"))
```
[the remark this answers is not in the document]
KiraNamida: It's a mismatch, can't recall if this is the name being wrong of the version being wrong but I suspect they'll be something between local and the pipeline that could differ here as there seemingly always is when we mess with timezones so will probably need to push to see what happens to be sure.

PR #1212, opened 2025-08-26
Update Preview settings to overwrite dev test testing in higher envs

KiraNamida:
Preview template is incorrectly serving the new version of the file as it's not overridden here so pulling the default values which are ahead of production release.


andrewl (approved):
lgtm

PR #1220, opened 2025-09-29
MRD-2907 Fix OWASP dependency check

ppmg:
OWASP dependency checks were failing due to OSS Index (which OWASP uses by
default) now requiring an API token to be used. The HMPPS gradle plug-in has
been updated to allow setting the token. The OWASP GHA workflow (which we
already use has also been updated accordingly to pass through the
username/token values when set as secrets, and the secrets have been set up
globally for MoJ projects, so just updating the plug-in version should fix the
OWASP issue for us.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,7 +1,7 @@
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 
 plugins {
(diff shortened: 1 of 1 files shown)


fincardy-moj (approved):
LGTM 👍

PR #1247, opened 2025-12-04
MRD-2945 Migrate from postman to intellij

fincardy-moj:
ITT: Fin goes a bit mad with the postman migration.

Sooo, whilst doing the postman migration I noticed that there's a tab on IntelliJ which works out all the endpoints mentioned in the code. So I've basically run through that and added all of them instead of just adding the postman ones. There'll be ones in here which aren't all that useful/relevant, but I figured if we're doing an overhaul we may as well try and cover as much as possible.

For the most part, it's split up into files based on the first URL segment - so if the request goes to `/recommendation/whatever` it'll be in the `recommendations.http` file. 

This method worked quite nicely, until I got the fiddly bits, which I've split up into their controller/domain - so, for example, the 3 endpoints under the PrisonAPIController are all in `prison-api`. 

More than happy for a discussion over this, as there's still a lot of this API which IU haven't touched and may make more sense to group together. I just want to ensure we're all getting what we want :)


ppmg (changes requested):

http/requests/arns-api.http
ppmg, line 1: Not sure if there's some criteria I'm missing regarding whether to include a consumes endpoint or not, so just in case I'll point out endpoints we consume not included in this file:
`/risks/crn/$crn/risk-management-plan`
and
`/risks/crn/$crn/fulltext`

Both are consumed by `ArnApiClient` if you need any more details.

http/http-client.env.json
```diff
@@ -26,7 +29,8 @@
           "Client Secret": "{{client-secret}}",
           "Auth URL": "https://sign-in-dev.hmpps.service.justice.gov.uk/auth/oauth/authorize",
           "Token URL": "https://sign-in-dev.hmpps.service.justice.gov.uk/auth/oauth/token",
-          "Client Credentials": "basic"
+          "Client Credentials": "basic",
+          "Redirect URL": "http://localhost:3000"
```
ppmg, line 33: Is this the correct redirect for dev?

fincardy-moj: Yeah this is because we're looking to use IntelliJ as "the thing that requests the token" (I'm sure there's a fancy OAuth word for this, but I can't remember it) so the dev auth instance will return the token to localhost:3000 which allows IntelliJ to pick it up

http/requests/ppud.http
```diff
@@ -0,0 +1,211 @@
+### Calls PPUD Automation service for a search
+POST {{host}}/ppud/details/{{ppudOffenderId}}
+Authorization: Bearer {{$auth.token("token")}}
+
+### Calls PPUD Automation service to create an offender
+POST {{host}}/ppud/offender
+Authorization: Bearer {{$auth.token("token")}}
+Content-Type: application/json
+
+{
```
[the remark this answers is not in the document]
fincardy-moj: Sorry, didn't reply to this - I did export the postman stuff during the FTR56 work to make sure I had a copy before the trial thingy ended, but checked those files and couldn't find any pre-filled requests so I suspect I've missed something on the export to get all those fields.

I have, however, updated the PPUD file so it's now set up to work off the private file which is in 1Password, it can then be tweaked for each dev from there.

http/requests/cases.http
```diff
@@ -0,0 +1,47 @@
...
+### Get details of the licence conditions on a case from CVL
+GET {{host}}/cases/{{crn}}/licence-conditions-cvl
+Authorization: Bearer {{$auth.token("token")}}
+
+### Get details of the licence conditions on a case
+GET {{host}}/cases/{{crn}}/licence-conditions/v2
+Authorization: Bearer {{$auth.token("token")}}
+
+### Get all details of a case contact history
+GET {{host}}/cases/{{crn}}/contact-history?query={{$random.alphanumeric(8)}}&from={{$random.alphanumeric(8)}}&to={{$random.alphanumeric(8)}}&type={{$placeholder}}&includeSystemGenerated=false
```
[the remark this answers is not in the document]
ppmg: The answer to this was that we just provide the minimal and maximal versions and leave it up to the developer to adjust as they need each time, as IntelliJ has a more scratchpad approach to this.

ppmg (approved):
LGTM!

PR #1256, opened 2026-01-30
MRD-2987 Retrieve correct sentence expiry dates

ppmg:
The expiry dates included in the sentences we get from the Prison API's
sentences-and-offences endpoint are inaccurate and in many cases no longer set.
Instead, we take the dates given in the sentenceDetail endpoint, which have
been calculated by the Calculate Release Dates service (which has the correct
calculations). We get a date per booking ID, as the calculation is done for a
whole "sentence calculation envelope", i.e. a whole sequence of connected
sentences (consecutive and concurrent). This means the date we get is the end
date of the whole sequence, which we then use to override whatever date we had
received from the sentences-and-offences endpoint. We do this to every sentence
in that sequence, even if the individual sentence's start date and length don't
add up to that end date, as we are only concerned with the overall end date.

Various refactors have been included here to simplify classes and testing.


fincardy-moj (dismissed):
LGTM!

fincardy-moj (approved):
LGTM

PR #1259, opened 2026-02-10
MRD-3023 Add Flipt for feature flags

ppmg:
We add support for using Flipt to control feature flags in the API service
here. As an immediate use, we also switch the document template version
selection to use Flipt flags rather than the previous config-based approach,
allowing us to change the values more easily if needed (e.g. to test in dev
or to adjust target release dates of new templates if the release date is
changed after we've added the template to the service).


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/documenttemplate/TemplateRetrievalService.kt
```diff
@@ -4,57 +4,56 @@ import org.slf4j.LoggerFactory
...
-        log.error("Recommendation identified with future created date: $ppDocumentCreated - Current date/time: $currentDateTime")
-        currentDateTime
-      } else {
-        ppDocumentCreated
-      }
+  private fun retrieveTemplateVersion(
+    documentType: DocumentType,
+    dateTimeOfFirstDownload: ZonedDateTime?,
+  ): TemplateVersion = when (documentType) {
+    DocumentType.PART_A_DOCUMENT, DocumentType.PREVIEW_PART_A_DOCUMENT -> templateVersionRetrievalService.partATemplateVersion(
```
sureshmoj, line 37: is this runtime error or startup error?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/documenttemplate/TemplateVersionRetrievalService.kt
```diff
@@ -0,0 +1,51 @@
...
+) {
+
+  companion object {
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
+
+  fun partATemplateVersion(dateTimeOfFirstDownload: ZonedDateTime?): PartATemplateVersion {
+    val currentDateTime = ZonedDateTime.now(ZoneId.of("UTC"))
+    val templateTargetDate =
+      if (dateTimeOfFirstDownload === null) {
```
sureshmoj, line 29: should this be == ?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/documenttemplate/TemplateRetrievalService.kt
ppmg: Sorry, I don't follow your question here. What error are you referring to?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/documenttemplate/TemplateVersionRetrievalService.kt
ppmg: Doesn't matter when it comes to null equality in Kotlin. See https://kotlinlang.org/docs/equality.html#structural-equality - see the first line after the first snippet of runnable code.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/documenttemplate/TemplateRetrievalService.kt
sureshmoj: Sorry, my comment is probably a bit outdated now, but I think it was referring to:

`throw DocumentTemplateNotFoundException("No Part A template version found (received null version)")`

Just wanted to check—if this happens, does it fail during service startup (causing the deployment to fail), or is it a runtime error that would only fail a particular request once the service is up and running?
or is that does not matter at all ?

But I think your comment above answers query, so we are good I think

PR #1292, opened 2026-04-13
MRD-3165 fix mappa review status on task list

fincardy-moj:
The bug was basically whenever you review "Personal Details" on the task list it was losing the "reviewed" status of the MAPPA information. This was because the personOnProbation refresh was removing the property saying the MAPPA info had been reviewed, this PR keeps that property intact.


sureshmoj (approved):
LGTM 👍

PR #1295, opened 2026-04-24
MRD-2987 Fix compilation error

ppmg:
Merges crossed each other which didn't conflict in mergin terms but
caused compilation issues. Fixed here.


fincardy-moj (approved):
LGTM!

PR #1296, opened 2026-04-24
MRD-3056: Use croNumber from NOMIS when croNumber not available in Delius

bhanugnosis:
MRD-3056: Use croNumber from NOMIS when croNumber not available in Delius

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapperTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapperTest.kt
@@ -48,6 +48,8 @@ import uk.gov.justice.digital.hmpps.makerecalldecisionapi.domain.makerecalldecis
 import uk.gov.justice.digital.hmpps.makerecalldecisionapi.domain.makerecalldecisions.recommendation.VictimsInContactScheme
 import uk.gov.justice.digital.hmpps.makerecalldecisionapi.domain.makerecalldecisions.recommendation.WhoCompletedPartA
 import uk.gov.justice.digital.hmpps.makerecalldecisionapi.domain.makerecalldecisions.recommendation.YesNoNotApplicableOptions
+import uk.gov.justice.digital.hmpps.makerecalldecisionapi.domain.makerecalldecisions.recommendation.personOnProbation
+import uk.gov.justice.digital.hmpps.makerecalldecisionapi.domain.makerecalldecisions.recommendation.prisonOffender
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapperTest.kt
```diff
@@ -188,6 +190,55 @@ class PartADocumentMapperTest {
...
+
+
+  @Test
+  fun `set croNumber from personOnProbation (Delius) when croNumber is present `() {
+    val nomisCro = "nomisCro"
+    runTest {
+      given(regionService.getRegionName(null))
+        .willReturn("")
+      val recommendation = RecommendationResponse(
+        personOnProbation = personOnProbation(croNumber = nomisCro).toPersonOnProbationDto()
```
ppmg: Could you also explicitly set a prisonOffender value here with a non-null CRO so we can see that prisonOnProbation has priority, please?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapperTest.kt
```diff
@@ -188,6 +190,55 @@ class PartADocumentMapperTest {
...
+    }
+  }
+
+  @Test
+  fun `set croNumber to null when croNumber not present in personOnProbation (Delius) and prisonOffender (NOMIS) `() {
+    runTest {
+      given(regionService.getRegionName(null))
+        .willReturn("")
+      val recommendation = RecommendationResponse(
+        personOnProbation = personOnProbation(croNumber = null).toPersonOnProbationDto()
```
ppmg: Could you explicitly set a prisonOffender with a null CRO here (instead of a null prisonOffender, as it does now), please?

ppmg (changes requested):

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapperTest.kt
bhanugnosis: Done

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapperTest.kt
bhanugnosis: Done

ppmg (approved):
LGTM!

PR #1299, opened 2026-05-11
MRD-2780 Migrate to GitHub Actions

ppmg:
This set of changes migrates the repo away from CircleCI and onto GitHub
Actions.

Notable details of these changes include:
- The e2e local tests are run in parallel jobs. See the
  hmpps-assess-risks-and-needs-github-actions repo for details on how the
  splitting and parallelisation is done. This is also being done in the e2e
  repo.
- The snyk security scan has been added. This replaces the trivy scan, which
  had to be removed a few weeks back due to security issues in trivy. The snyk
  one has been added using the snyk-add.sh script in the hmpps-github-actions
  repo.
- SonarQube is already due to lose support at some point, as was announced in
  Slack a while back, so we've decided to use this migration to remove it
  altogether. With this, we also remove jacoco, which was generating the data
  for SonarQube. We haven't been doing separate checks on the code coverage
  data jacoco produces, so there's no point in keeping it. If we do need it in
  the future we can always re-add it.
- The -Dkotlin.compiler.execution.strategy=in-process and
  -Dorg.gradle.workers.max=1 java options that existed in the CircleCI validate
  job were originally added for sonar, hence why we aren't preserving them.
- There are other java options that haven't been preserved. It is unclear how
  necessary there were, but running the new workflows without them seems to
  work fine and in similar times as the old ones, so will leave them out. Can
  add them back if needed in the future.
- The resource_class option in the validate job was to handle the switch from
  H2 to running postgres from a Docker container. Might be worth checking if
  this is still needed when we move to testcontainers with MRD-2840.
- The run-e2e-tests.sh script has been removed. It was originally used for e2e
  test runs against dev and preprod, but those jobs have been commented out
  from this repo for a long time now. If we do enable e2e tests against dev for
  this repo, we will likely do so in alignment with how the UI and e2e-tests
  repos do it anyway.


.circleci/config.yml.bak.20250918_093110
```diff
@@ -1,3 +1,38 @@
+pablo.glenn@MJ004118 make-recall-decision-api % /bin/bash -c "$(curl -fsSL https://github.com/ministryofjustice/hmpps-github-actions/raw/refs/heads/main/migrate-repo.sh)"
```
fincardy-moj, line 1: Is this line supposed to be here? Don't think it matters, but a weird sight!

fincardy-moj (changes requested):

.circleci/config.yml.bak.20250918_093110
ppmg: Yeah, I just copied over the contents of my command line run - the stuff after this is the output of executing the command above. Just to have a bit of a history of how some of this migration was done.

fincardy-moj: Ah cool, that makes sense!

fincardy-moj (dismissed):
LGTM!

fincardy-moj (approved):
LGTM!

PR #1301, opened 2026-05-22
Mrd 3158 upgrade to spring boot 4

fincardy-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/ppud/PpudAutomationApiClient.kt
```diff
@@ -38,7 +38,7 @@ import java.util.concurrent.TimeoutException
 
 class PpudAutomationApiClient(
   private val webClient: WebClient,
-  @Value("\${ppud-automation.client.timeout}") private val ppudAutomationTimeout: Long,
+  @param:Value("\${ppud-automation.client.timeout}") private val ppudAutomationTimeout: Long,
```
fincardy-moj, line 41: This `@param` is to remove the following warnings on compile:

```
This annotation is currently applied to the value parameter only, but in the future it will also be applied to field.
- To opt in to applying to both value parameter and field, add '-Xannotation-default-target=param-property' to your compiler arguments.
- To keep applying to the value parameter only, use the '@param:' annotation target.

See https://youtrack.jetbrains.com/issue/KT-73255 for more details.
```

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/config/CustomOAuth2ClientCredentialsGrantRequestEntityConverter.kt
```diff
@@ -1,22 +0,0 @@
-package uk.gov.justice.digital.hmpps.makerecalldecisionapi.config
-
```
fincardy-moj, line 2: This method of adding the parameters to the token grant request has been replaced here: 
https://github.com/ministryofjustice/make-recall-decision-api/pull/1301/changes#diff-877f19ae507f3f0ce92c588afd8c75f9e96154cc434f7cf20a43a790f3193bd7L76

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/config/WebConfig.kt
```diff
@@ -2,17 +2,11 @@ package uk.gov.justice.digital.hmpps.makerecalldecisionapi.config
...
 
 @Configuration
 class WebConfig : WebMvcConfigurer {
   override fun addInterceptors(registry: InterceptorRegistry) {
     registry.addInterceptor(SentryContextAppender())
   }
-
-  override fun configurePathMatch(configurer: PathMatchConfigurer) {
-    // Use this until the UI is updated to no longer use trailing slashes
-    configurer.setUseTrailingSlashMatch(true)
```
fincardy-moj, line 16: This was deprecated and required the use of `UrlHandlerFilter` to get the same behaviour, but given our UI only adds a trailing slash to the `[PATCH] /recommendation/<recommendationId>` endpoint (See here: https://github.com/ministryofjustice/make-recall-decision-ui/blob/f8bd8a88dde58ac0566e858729ab4ab642929180/server/data/makeDecisionApiClient.ts#L371) and the `UrlHandlerFilter` method of doing it seems to require specifying paths anyways, I figured I'd just add the `/` version to the `@PatchMapping` annotation alongside the non-trailing slash version here: https://github.com/ministryofjustice/make-recall-decision-api/pull/1301/changes#diff-5c2fae76737e8d968e433a08221fe928229b2ae6b55ab050372b5c56847661f8L130

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/recommendation/AdditionalLicenceConditionOption.kt
```diff
@@ -1,9 +1,11 @@
...
+import java.io.Serializable
+
 data class AdditionalLicenceConditionOption(
   val subCatCode: String? = null,
   val mainCatCode: String? = null,
   val title: String? = null,
   val details: String? = null,
   val note: String? = null,
-)
+) : Serializable
```
fincardy-moj, line 11: _alllllllll_ the classes under the RecommendationEntity must now have `Serializable`

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/prisonapi/PrisonerApiService.kt
```diff
@@ -80,7 +80,7 @@ internal class PrisonerApiService(
         val lastDateOutOfPrison =
           prisonPeriod.movementDates.map { it.dateOutOfPrison }.filter { it != null }
             .maxWithOrNull(Comparator.naturalOrder())
-        val movement = prisonPeriod.movementDates.find { it.dateOutOfPrison === lastDateOutOfPrison }
+        val movement = prisonPeriod.movementDates.find { it.dateOutOfPrison == lastDateOutOfPrison }
```
fincardy-moj, line 83: LocalDateTime instances can be different objects, so a looser equality was needed here.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/MrdEventsEmitter.kt
```diff
@@ -23,10 +23,10 @@ class MrdEventsEmitter(
   objectMapper: ObjectMapper,
   customTelemetryClient: TelemetryClient,
 ) {
-  private lateinit var domainEventTopicSnsClient: SnsAsyncClient
-  private lateinit var topicArn: String
-  private lateinit var objectMapper: ObjectMapper
-  private lateinit var telemetryClient: TelemetryClient
+  private var domainEventTopicSnsClient: SnsAsyncClient
```
fincardy-moj, line 26: This was another compile warning just pointing out that all of these are set in the constructor so `lateinit` isn't needed here.

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/CaseOverviewControllerTest.kt
```diff
@@ -247,7 +248,10 @@ class CaseOverviewControllerTest(
       userAccessAllowed(crn)
       overviewResponse(crn, delaySeconds = nDeliusTimeout + 2)
 
-      webTestClient.get()
+      webTestClient.mutate()
```
fincardy-moj, line 251: The webTestClient is more strict with its default timeout in newer versions, so we need to set a timeout that's longer than the expected nDeliusTimeout (or oasysTimeout or whatever timeout we're testing)

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/integration/controller/SubjectAccessRequestIntegrationTest.kt
```diff
@@ -44,11 +45,13 @@ class SubjectAccessRequestIntegrationTest : IntegrationTestBase() {
     @Bean
     open fun sarIntegrationTestHelper(
       jwtAuthorisationHelper: JwtAuthorisationHelper,
+      objectMapper: ObjectMapper,
```
fincardy-moj, line 48: Had to upgrade to the spring boot 4-compatible version of the SAR library which expects an objectMapper

build.gradle.kts
```diff
@@ -54,32 +56,30 @@ dependencies {
...
-  implementation("org.flywaydb:flyway-core:11.1.1")
-  implementation("org.flywaydb:flyway-database-postgresql:11.1.1")
-  implementation("org.postgresql:postgresql:42.7.11")
+  implementation("org.springframework.boot:spring-boot-starter-flyway")
+  implementation("org.flywaydb:flyway-database-postgresql")
+  implementation("org.postgresql:postgresql:42.7.11") // hmpps-kotlin-spring-boot-starter pulls in 42.7.10 - should we remove this line here and leave it up to the starter?
 
-  implementation("io.sentry:sentry-spring-boot-starter-jakarta:7.20.0")
+  implementation("io.sentry:sentry-spring-boot-4:8.42.0")
   implementation("io.sentry:sentry-logback:7.20.0")
```
sureshmoj: should this **sentry-logback** be 8.42.0 as well? otherwise this will be off with **sentry-spring-boot** and may cause some errors or is that not an issue ?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/controller/AuthenticationFacade.kt
```diff
@@ -23,9 +23,11 @@ class AuthenticationFacade {
 
   val isSpoOrAco: Boolean
     get() {
-      val credentials = authentication.credentials as Jwt
+      val credentials = authentication?.credentials as Jwt
```
sureshmoj: minor : but can we do null safe check ? **authentication?.credentials as? Jwt ?**

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/service/PpcsService.kt
```diff
@@ -25,8 +25,10 @@ internal class PpcsService(
   fun search(crn: String): PpcsSearchResponse {
     log.info("ppcs searching for crn: " + crn)
 
+    val username = authenticationFacade.authentication?.name!!
```
sureshmoj: Using ?. then !! is contradictory and reads oddly, can we use ?  syntax instead

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/config/WebConfig.kt
sureshmoj: is this something to keep in mind if we see any 404 route errors in prod ? :)

src/main/resources/application.yml
```diff
@@ -82,8 +85,6 @@ spring:
     parser:
       allow-backslash-escaping-any-character: true
     date-format: "yyyy-MM-dd HH:mm:ss"
-    serialization:
-      WRITE_DATES_AS_TIMESTAMPS: false
```
sureshmoj, line 86: why removing this ?  is it false by default in spring boot 4 ? 
if not, the dates would serialize as epoch millis instead of ISO strings, isnt it ?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/DeliusClient.kt
```diff
@@ -370,7 +397,7 @@ class DeliusClient(
     )
   }
 
-  data class ContactTypeSummary(val code: String, val description: String, val total: Int)
+  data class ContactTypeSummary(val code: String, val description: String, @param:JsonAlias("hits") val total: Int)
```
sureshmoj: is the delius api changed to use **hits** here rather than **total**?  in that case if we replace **total** with **hits**, would that break things?

src/main/resources/application.yml
fincardy-moj: Yeah, Spring 4 moved to Jackson 3 by default and they've updated this setting:
https://spring.io/blog/2025/10/07/introducing-jackson-3-support-in-spring#adapt-to-the-new-default-settings

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/config/WebConfig.kt
fincardy-moj: Yep, definitely worth keeping in mind but I've been as vigilant as possible with these so can confirm that it's only this update recommendation which *could* have the trailing slash.

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/client/DeliusClient.kt
fincardy-moj: This was a good shout, turns out I'd got my wires crossed and it was actually the `deliusContactHistoryResponse` fixture having a misnamed field which the spring update was highlighting, so have updated that and removed this JSON alias

sureshmoj (dismissed):
LGTM 👍

ppmg (approved):
LGTM

PR #1312, opened 2026-06-17
Remove unnecessary taq file

ppmg:
The taq file looks like it was accidentally committed, so we are removing it.


fincardy-moj (approved):
👋

PR #1314, opened 2026-06-24
MRD-3225: Pass sentencedAsYouth field to PPUD

sureshmoj:


ppmg (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/PpudCreateOffenderRequest.kt
```diff
@@ -19,6 +19,7 @@ data class PpudCreateOffenderRequest(
   val mappaLevel: String?,
   val nomsId: String?,
   val prisonNumber: String?,
+  val sentencedAsYouth: String? = null,
```
ppmg: Is it worth having an enum defined for this instead of using `String?` so we know what the possible values are?

sureshmoj: fixed, but existing fields like custodyType also use string type for some reason.

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/ppud/PpudCreateOffenderRequestTestBuilder.kt
```diff
@@ -26,7 +27,7 @@ internal fun ppudCreateOffenderRequest(
   mappaLevel: String? = randomString(),
   nomsId: String? = randomString(),
   prisonNumber: String? = randomString(),
-  sentencedAsYouth: String? = randomString(),
+  sentencedAsYouth: SentencedAsYouth? = SentencedAsYouth.No,
```
ppmg: Small point: could you use `randomEnum<SentencedAsYouth>()` instead, please?

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/PpudCreateOrUpdateSentenceRequestTestBuilder.kt
```diff
@@ -24,7 +24,7 @@ fun ppudCreateOrUpdateSentenceRequest(
   sentenceExpiryDate: LocalDate? = randomLocalDate(),
   sentencingCourt: String = randomString(),
   sentencedUnder: String = randomString(),
-  sentencedAsYouth: String? = randomString(),
+  sentencedAsYouth: SentencedAsYouth? = SentencedAsYouth.No,
```
ppmg: Small point: could you use `randomEnum<SentencedAsYouth>()` instead, please?

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/PpudCreateOffenderRequest.kt
ppmg: I think that might be because we get the custody types from PPUD in the first place and the user is given a dropdown list from which to pick, but arguably we could define enums for `custodyType` and the like too. In this case we pick the value automatically based on `sentenceGroup`, so no need to offer options retrieved from PPUD.

The suggestion of using boolean instead (in the UI PR) would simplify things here and in UI, but I don't think we'd ever get rid of the risk of becoming out of sync with PPUD if they ever decide to change the dropdown texts for this new field, so not sure if one approach is better than another.

ppmg (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/PpudCreateOffenderRequest.kt
sureshmoj: yeah, agreed, lets keep the enum

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/PpudCreateOrUpdateSentenceRequestTestBuilder.kt
sureshmoj: fixed

src/test/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/domain/makerecalldecisions/ppud/PpudCreateOffenderRequestTestBuilder.kt
sureshmoj: fixed

ppmg (approved):
LGTM

PR #1321, opened 2026-07-07
Fix pipeline

ppmg:
The e2e_local_tests step had been missed out when setting the build step's
requirements. We don't want the image built until we confirm all tests are
running successfully (and more importantly because we don't want to deploy into
dev or be able to deploy into prod without the local e2e tests passing).


sureshmoj (approved):
LGTM 👍

PR #1325, opened 2026-07-13
fix API cves

fincardy-moj:


build.gradle.kts
```diff
@@ -58,7 +58,7 @@ dependencies {
 
   implementation("org.springframework.boot:spring-boot-starter-flyway")
   implementation("org.flywaydb:flyway-database-postgresql")
-  implementation("org.postgresql:postgresql:42.7.11") // hmpps-kotlin-spring-boot-starter pulls in 42.7.10 - should we remove this line here and leave it up to the starter?
+  implementation("org.postgresql:postgresql:42.7.12")
```
ppmg, line 61: Note to selves for the future: we removed this comment because we could find the referred starter ever having pulled in postgresql and it made no sense to keep the comment.

ppmg (approved):
LGTM

PR #1349, opened 2026-07-16
Pin GHA actions

ppmg:
Calls to other GHA actions should be pinned with a SHA to prevent supply chain
attacks. HMPPS guidelines tell us we only need to do this with actions;
workflows can still use the @ version specification method.


fincardy-moj (approved):
LGTM!

PR #1355, opened 2026-07-20
MRD-3244 Add new Part A template

ppmg:
For the FTR56 sentence conviction development.


src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapper.kt
```diff
@@ -126,6 +127,10 @@ internal class PartADocumentMapper(
         recommendation.hasBeenChargedWithTerroristOrStateThreatOffence,
         recommendation,
       ),
+      isChargedAndConvictedWithOffence = generateExclusionCriteriaAnswer(
+        recommendation.isRecalledOnNewChargedOrConvictedOffence?.selected,
```
sureshmoj, line 131: minor: but do you need to pass this arg as you are already passing the `recommendation` so you can compute it in the calling function ?

sureshmoj (approved):
lgtm 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/makerecalldecisionapi/documentmapper/PartADocumentMapper.kt
ppmg: Yes and no. Yes, I could definitely change it. However, I'm following the pattern of other functions with the same name in this file. If I were to reduce them all to have a single parameter, I'd need to change the function names (or at least one of them), as you can't overload a method with the exact same definition. Ultimately I went with the overload approach, but happy to change it if you think it's confusing this way?

sureshmoj: not a biggie  - please merge

PR #1357, opened 2026-07-21
MRD-3024 mock the flipt endpoint

fincardy-moj:
This PR will create a mock endpoint for a Flipt server, for use with "local" E2E tests

### DIFF
--- a/docker-compose-e2e.yml
+++ b/docker-compose-e2e.yml
@@ -143,6 +143,19 @@ services:
     volumes:
       - ./fake-prison-api:/home/wiremock
 
+  fake-flipt-api:
+    image: wiremock/wiremock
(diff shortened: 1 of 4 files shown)


docker-compose-e2e.yml
```diff
@@ -143,6 +143,19 @@ services:
     volumes:
       - ./fake-prison-api:/home/wiremock
 
+  fake-flipt-api:
```
ppmg, line 146: Would it not make more sense to have this in the UI docker-compose file? Or is there some other reason that requires it to be here?

ppmg: (this and the rest of the changes in the PR, I mean)

